PR #63, opened 2020-09-18
[cli, exec, lib] Remove effective-klio-job.yaml

DanSimon:
Two main important changes in this PR:

### Config hand-off from CLI to exec

When CLI `run` invokes exec `run`, it now generates a "materialized" config file on the fly and hands it off to exec (purposely using "materialized" instead of "effective" to eliminate any possible confusion with the existing effective config).  This file includes any overrides and template parameters applied, so exec doesn't have to worry about any of that and just reads the generated file as-is.  

This basically solves the problem of how to propagate config overrides from CLI to exec.

Note: since we have since decided to add all config parsing support to exec to support Luigi (which will bypass CLI and call exec directly), an alternative approach could have been to simply pass-on all overrides as command-line parameters when CLI invokes exec.  However, personally I think this is still a cleaner solution but I'm happy to discuss this further.

### Exec config pickling

Once exec has loaded config, it uses `RunConfig` to set a main-session global variable to the config object.  When the beam pipeline starts, Beam itself will automatically pickle all main-session vars and include them on Dataflow workers (and of course in direct-runner too).  This eliminates the need for workers to load `effective-klio-job.yaml` currently included in docker images and python packages.

This solves two problems:
1.  Since `effective-klio-job.yaml` was included during `klio image build`, there could be a mismatch between the packaged config file and the actual file when users run `klio job run`.  We currently warn about this, but it still leads to confusion.  Plus, right now making any config change requires re-running `klio image build`.
2.  Similarly, this also meant that we had no easy way to support config overrides, since even though klio CLI can parse the overrides and modify its config, klio exec still just loaded the packaged file.  On dataflow workers, this could happen before any code is executed since some decorators read config and thus config is needed as code using the decorators is imported.

## Checklist for PR author(s)

### DIFF
--- a/cli/src/klio_cli/cli.py
+++ b/cli/src/klio_cli/cli.py
@@ -139,9 +139,6 @@ def run_job(klio_config, config_meta, **kwargs):
         config_meta.job_dir, kwargs.get("image_tag")
     )
     image_tag = kwargs.get("image_tag") or git_sha
-    if config_meta.config_file:
-        basename = os.path.basename(config_meta.config_file)
-        image_tag = "{}-{}".format(image_tag, basename)
 
     runtime_config = DockerRuntimeConfig(
         image_tag=image_tag,
--- a/exec/src/klio_exec/cli.py
+++ b/exec/src/klio_exec/cli.py
@@ -14,19 +14,21 @@
 #
 
 import collections
-import hashlib
 import logging
 import os
 
 import click
 import yaml
 
+# WARNING: be careful with what is imported here.  Any decorators that read
+# config cannot be imported here (even transitively) since they will attempt to
+# read the config object before it exists.
+from klio.transforms import core as klio_transforms_core
 from klio_core import config
 from klio_core import options as core_options
 
 from klio_exec import options
 from klio_exec.commands import audit
-from klio_exec.commands import run
 from klio_exec.commands import stop
 
 
@@ -54,25 +56,9 @@ def _get_config(config_path):
         raise SystemExit(1)
 
 
+# TODO : remove this when the call in spotify-klio is removed
 def _compare_runtime_to_buildtime_config(runtime_config_path):
-    buildtime_config_path = "/usr/src/config/.effective-klio-job.yaml"
-    # to maintain backwards compatibility with klio lib < 0.0.18
-    if not os.path.exists(buildtime_config_path):
-        return True
-
-    # use the same hash algo as docker
-    runtime_hasher = hashlib.sha256()
-    buildtime_hasher = hashlib.sha256()
-
-    with open(runtime_config_path, "rb") as rt_conf:
-        buff = rt_conf.read()
-        runtime_hasher.update(buff)
-
-    with open(buildtime_config_path, "rb") as bt_conf:
-        buff = bt_conf.read()
-        buildtime_hasher.update(buff)
-
-    return buildtime_hasher.hexdigest() == runtime_hasher.hexdigest()
+    return True
 
 
 @main.command("run")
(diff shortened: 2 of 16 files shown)


fallonchen (commented):
Some initial thoughts - 
* This seems like a safe approach, but I'm taking a little extra time to look at how/when Beam (and the Portability API) does pickling and loading of main session variables for my own understanding, so the review will be a bit slow, sorry!
* I'm curious about the parameters for the test you ran for this and whether or not they impacted the accessibility of the main session variables - Was the job batch? streaming? What version of Beam? Did it use FnAPI or setup.py? Was the workload threaded? Was it a basic `beam.ParDo(SomeDoFn())` type of pipeline or something more complicated? 
* We'd have to follow this with a similar change in internal `klio-exec`, right? 
* Could you amend the commit message to something more descriptive?

DanSimon:
@fallonchen I ran a few different tests, both with streaming and batch jobs on both direct-runner and dataflow.  The large-scale test I ran was a batch job that read from a ~1B row BQ table.  The job itself didn't really do anything except attempt to read a config value in the `process` method.  I believe I used FnApi in my tests, but you bring up a good point to try this with setup.py.  I will double check and run some more tests.

As far as I can tell, main session variable unpickling happens before any user code is executed.  Some of the transform metaclasses read config and are evaluated when the code is imported.  They don't seem to have any problem reading config this way.

exec/src/klio_exec/cli.py
```diff
@@ -84,23 +70,19 @@ def run_pipeline(image_tag, direct_runner, update, config_file, blocking):
...
         config_data["pipeline_options"]["runner"] = "direct"
 
     job_name = config_data["job_name"]
     conf_obj = config.KlioConfig(config_data)
+
+    # RunConfig ensures config is pickled and sent to worker.  Note this
+    # depends on save_main_session being True
+    klio_transforms_core.RunConfig.set(conf_obj)
+
+    # This can only be imported after RunConfig is set
```
fallonchen, line 83: Can you expand on why here?

cli/src/klio_cli/cli.py
```diff
@@ -139,9 +139,6 @@ def run_job(klio_config, config_meta, **kwargs):
         config_meta.job_dir, kwargs.get("image_tag")
     )
     image_tag = kwargs.get("image_tag") or git_sha
-    if config_meta.config_file:
-        basename = os.path.basename(config_meta.config_file)
-        image_tag = "{}-{}".format(image_tag, basename)
```
econchick, line 144: The reason this was in here was to be able to tell which config the image was built with (for those using multiple configuration files). I'm not sure I understand the reasoning behind removing it?

core/src/klio_core/testing.py
```diff
@@ -0,0 +1,90 @@
+# Copyright 2020 Spotify AB
```
econchick, line 1: Are you intending that users can use objects from this module? If not, can we make it `_testing.py`?

If you are, I would want it `klio` since `klio-core` isn't intended for users (although I understand that screws up dependencies for the klio-cli).

econchick (commented):
This looks pretty good - just couple of questions.

cli/src/klio_cli/cli.py
DanSimon: hmm that's an interesting point.  I did this because I thought the reason it was included in the tag name was to make it easier to know which config file was included in the image as the `effectiive-klio-job.yaml`.  Since that's no longer used, there's no longer a need for that kind of indication.

I do agree though that it would be good to know what config was used when the image was built, however now that I think about it, adding the file name to the image tag would still result in some ambiguity because:
1.  You don't know what version of the file was used, only the name
2.  Now that we support overrides, it could be misleading if the user provides a CLI arg to `klio image build` that affects the build process

So I still kinda think this change makes sense, but maybe it would be a good idea to include a `build-config.yaml` or something in the image?

core/src/klio_core/testing.py
DanSimon: Oh yeah I don't think users need to use this, will change to `_testing.py`

cli/src/klio_cli/cli.py
econchick: Those are fair arguments. I think a `build-config.yaml` makes sense too. Perhaps a separate ticket, so we can get this PR through?

DanSimon: yeah I agree that can be addressed separately

PR #92, opened 2020-10-08
[site] Add initial landing page

econchick:
Adding landing page / micro site to be served by [GH pages](https://docs.github.com/en/free-pro-team@latest/github/working-with-github-pages) (which I'll set up after this gets merged). 

To view:

```sh
# w/i repo
$ git fetch origin
$ git checkout -b lynn/add-landing-site origin/lynn/add-landing-site
$ python -m http.server 8888  # no virtualenv needed
```

Then navigate to localhost:8888

I basically re-built the exported-from-Figma site with using Bootstrap. I'm not a front end person, so some things may be hacky. Open to suggestions!

## Checklist for PR author(s)


microsite/index.html
```diff
@@ -0,0 +1,100 @@
...
+              <a class="nav-link" href="https://klio.readthedocs.io"><button class="gradient">Get Started</button></a>
+            </li>
+          </ul>
+        </div>
+      </nav>
+    </header>
+
+    <main role="main" class="flex-shrink-0">
+      <div class="shove-down"></div>
+      <div class="container mt-5">
```
fallonchen, line 40: Wondering if it's possible to increase the height of this and/or add another shove-down div to increase the space between the wave image and the text here. Fine with not doing it if it's more complicated than that!

fallonchen (dismissed):
🎉   

I think the one difference I see between the designer's version and this is that the space between the wave and the blurb is a bit smaller than the designer's version (see comment).  Looks good either way, so it's up to you if you want to change it more to match.

microsite/index.html
econchick: @fallonchen That can be done, sure! I wasn't too concerned with keeping things _exact_ but if you think it looks better / is more readable, I'm happy to change it!

PR #99, opened 2020-10-16
Move with_klio_config to core

shireen-bean:
## Checklist for PR author(s)


DanSimon (dismissed):
couple very small things, but overall looks great!   Deleting all the duplicate options was a very nice unintended side effect 👍

core/src/klio_core/options.py
```diff
@@ -0,0 +1,265 @@
+# Copyright 2020 Spotify AB
```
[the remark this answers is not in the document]
shireen-bean: There's no code / logic changes but it takes the overlapping options from cli and exec and unites them here. So it's not a direct replica of cli or exec but as far as individual functions, those are unchanged, just moved

core/src/klio_core/options.py
```diff
@@ -127,25 +139,127 @@ def wrapper(func):
...
+            "elements process."
         ),
     )(func)
+
+
+def per_element(func):
+    return click.option(
+        "--per-element",
+        default=False,
+        show_default=True,
```
econchick, line 246: In the cli version of this option, it's `show_default=True` (as it is here); in the [exec version of this option](https://github.com/spotify/klio/blob/master/exec/src/klio_exec/options.py#L226), it's `show_default="True"` (`bool` vs `str`). They are two different behaviors: the first will show that the default is set to `False` as declared right above it; the second will show the default is "True" because it's a string. Also notice the [comment](https://github.com/spotify/klio/blob/master/exec/src/klio_exec/options.py#L226) for the exec option - it tells me that the logic for handling the option will actually be a little different than how the CLI will handle it. 

We should keep the original behavior, however it would be good to not have two `per_element` options that are otherwise quite similar. This can be achieved by defining an inner function, similar to how it's done with [`job_name`](https://github.com/spotify/klio/blob/master/cli/src/klio_cli/options.py#L197-L205):

```py
# in core/src/klio_core/options.py
def per_element(*args, **kwargs):
    def wrapper(func):
        return click.option( 
            "--per-element",
            default=False,
            is_flag=True,
            cls=MutuallyExclusiveOption,
            mutually_exclusive=["maximum"],
            help="Print memory usage per line for each input element processed.",
            **kwargs,  # added (and no more show_default keyword arg)
        )(func)

    if args:
        return wrapper(args[0])
    return wrapper

# in exec/src/klio_exec/cli.py
# for all the commands that use the per_element option:
@core_options.per_element(show_default="True")
...

# in cli/src/klio_cli/cli.py
# for all the commands that use the per_element option:
@core_options.per_element(show_default=True)
...
```

core/src/klio_core/options.py
```diff
@@ -127,25 +139,127 @@ def wrapper(func):
...
+        "--show-logs",
+        default=False,
+        show_default=True,
+        is_flag=True,
+        help="Show a job's logs while profiling.",
+    )(func)
+
+
+#####
+# options for `klioexec profile memory
```
econchick, line 174: The comments throughout this module are not consistent in terms of what they're for - for instance, these look specific to `klioexec` when it's also used for the klio cli. Maybe just generalize them to say "options for \`\<cmd> profile memory\`" (or whatever it's use for).

core/src/klio_core/options.py
```diff
@@ -67,19 +67,30 @@ def handle_parse_result(self, ctx, opts, args):
         )
 
 
-def _verify_gcs_uri(ctx, param, value):
-    if value and not value.startswith("gs://"):
-        raise click.BadParameter(
-            "Unsupported location type. Please provide a GCS location with "
-            "the `gs://` prefix."
-        )
+def override(func):
```
econchick, line 70: In the future, please maintain [atomic commits](https://www.freshconsulting.com/atomic-commits/) - this function and the `template` function are not relevant to the purpose of this particular commit (c8248988).

core/src/klio_core/options.py
```diff
@@ -0,0 +1,151 @@
...
+        )(func)
+
+    # allows @options.foo to be used without parens (i.e. no need to do
+    # `@options.foo()`) when there are no args/kwargs provided
+    if args:
+        return wrapper(args[0])
+    return wrapper
+
+
+def config_file(*args, **kwargs):
```
econchick, line 118: I see you don't actually end up using this in exec. In the next commit you still have the option defined in `klio_exec/options.py` it looks like (and not using `@core_options.config_file` in `klio_exec/cli.py`). I would not move it over to klio_core if it's not being used (they're different enough to keep it separate, for now at least).

core/src/klio_core/options.py
```diff
@@ -0,0 +1,265 @@
...
+#####
+def iterations(func):
+    return click.option(
+        "-n",
+        "--iterations",
+        default=10,
+        show_default=True,
+        type=int,
+        help="Number of times to execute each entity ID provided.",
+    )(func)
```
econchick, line 265: If you want, you can also add [`image_tag`](https://github.com/spotify/klio/blob/master/cli/src/klio_cli/options.py#L154-L160) ([klio-exec definition](https://github.com/spotify/klio/blob/master/exec/src/klio_exec/options.py#L72)). It would need to be define / used similarly to `per_element`:

```py
# in core/src/klio_core/options.py
def image_tag(*args, **kwargs):
    def wrapper(func):
        return click.option( 
            "--image-tag",
            help="Docker image tag to use"
            **kwargs
        )(func)

    if args:
        return wrapper(args[0])
    return wrapper

# in exec/src/klio_exec/cli.py
# for all the commands that use the image_tag option:
@core_options.image_tag
...

# in cli/src/klio_cli/cli.py
# for all the commands that use the image_tag option:
@core_options.image_tag(default=None, show_default="``git-sha[dirty?]``")
...
```

econchick (dismissed):
Please don't forget to update the API docs too, specifically for `klio-core` (as `klio-cli` just currently documents the CLI itself, and there are no API docs for [klio-exec](https://docs.klio.io/en/latest/reference/executor/index.html) yet). 

Changes are needed for [the summary page](https://github.com/spotify/klio/blob/master/docs/src/reference/core/index.rst) and the [utils page](https://docs.klio.io/en/latest/reference/core/api/utils.html). Right now, the only documented util option you have is the `with_klio_config` function. So, you can either: only add that function to be included in both pages that need updating, or you can go above and beyond and write docstrings for the other functions/classes that have been added to the `klio_core.utils` module here, and then include them all in those two doc pages.

For now, I think it's fine to not create documentation for the options module in klio core, since none of them are really documented.

core/src/klio_core/options.py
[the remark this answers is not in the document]
shireen-bean: So `with_klio_config` option is dependent on cli options
```@options.override
    @options.template
    @options.job_dir
    @options.config_file```

So I had to move those over to core regardless. But i agree there was an extra step of moving some of the other overlaps into core from exec and config also

File: core/src/klio_core/options.py (hunk 5 above)
Comment (shireen-bean) [reply core/src/klio_core/options.py:118]: It's being used by klio_core as part of `with_klio_config`

File: core/setup.py [hunk 7]
```diff
@@ -86,6 +87,7 @@ def find_meta(meta):
     "docs": ["sphinx", "interrogate"],
     "tests": [
         "coverage",
+        "mock",  # for py2 - remove when dropping support
```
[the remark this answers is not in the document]
shireen-bean: from `test_warn_if_py2_job` which uses mock and was moved to klio core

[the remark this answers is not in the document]
econchick: It's only needed for py2 support (so obv it wasn't removed when we dropped py2 support 😬 ). For py3+ you should be able to replace `import mock` with `from unittest import mock`.

shireen-bean:
This PR has been split into two autonomous commits

DanSimon (approved):
👍

PR #111, opened 2020-10-30
[cli, core, exec, lib] unpin google lib versions, support beam 2.25

DanSimon:
* Removing most of the versions restriction on google lib dependencies.
* including testing with beam 2.25 (left 2.22 in for now but we can probably stop testing on it soon)
* removing use of `2020-resolver` in dependency check workflow because of pip issues

## Checklist for PR author(s)

### DIFF
--- a/exec/tox.ini
+++ b/exec/tox.ini
@@ -1,5 +1,5 @@
 [tox]
-envlist = {py36,py37,py38}-beam{22,23,24},docs,manifest,check-formatting,lint
+envlist = {py36,py37,py38}-beam{23,24,25},docs,manifest,check-formatting,lint
 
 [testenv]
 setenv =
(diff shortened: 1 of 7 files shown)


lib/tests/unit/transforms/test_core.py
```diff
@@ -168,6 +168,8 @@ def test_job_property(thread_local_ret, mocker, monkeypatch):
         mock_func.assert_not_called()
         assert klio_ns._thread_local.klio_job == ret_value
 
+    klio_ns._thread_local.klio_job = None
```
DanSimon, line 171: This is an odd one.  With beam 2.25, this test suddenly started causing tests in `test_helpers` to start failing.  Apparently the mock stuck around in `_thread_local` and other tests tried to use it 😬

fallonchen (approved):
:shipit:

exec/tox.ini
```diff
@@ -1,5 +1,5 @@
 [tox]
-envlist = {py36,py37,py38}-beam{22,23,24},docs,manifest,check-formatting,lint
+envlist = {py36,py37,py38}-beam{22,23,24,25},docs,manifest,check-formatting,lint
```
econchick, line 2: We only test the latest 3 beam versions

DanSimon: if we stop testing on 22 should we also then set the min version for beam to `>2.22`?  I left it in because in my other branch that fixed BQ reader, my first attempt to support beam 2.25 ended up breaking under 2.22, something I would have missed otherwise.

econchick: I wouldn't set the minimum beam version; we just only explicitly test the latest 3 beam releases to minimize our test matrix a bit (and we're a bit lucky that many of our current users update beam versions frequently).

PR #116, opened 2020-11-06
[docs] Add docs for maintainers on releasing

econchick:
Figured these might be helpful.

## Checklist for PR author(s)

### DIFF
--- a/RELEASING.rst
+++ b/RELEASING.rst
@@ -1,0 +1,360 @@
+Releasing a Klio Package
+========================
+
+Klio packages are manually released and uploaded to PyPI (versus via CI/CD).
+The following instructions assume you're an `approved maintainer <https://github.com/spotify/klio/blob/master/CODEOWNERS>`_ of Klio.
+
+
+*Credit: These instructions are heavily inspired by* |this write up|_.
+
+.. it's impossible to nest formatting - aka have a link inside an italicized line. So this is a hack from https://stackoverflow.com/questions/4743845/format-text-in-a-link-in-restructuredtext
+
+.. |this write up| replace:: *this write up*
+.. _this write up:  https://hynek.me/articles/sharing-your-labor-of-love-pypi-quick-and-dirty/
+
+.. _initial-release-setup:
+
+Initial Setup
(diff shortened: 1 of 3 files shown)


RELEASING.rst
```diff
@@ -0,0 +1,383 @@
...
+
+.. |this write up| replace:: *this write up*
+.. _this write up:  https://hynek.me/articles/sharing-your-labor-of-love-pypi-quick-and-dirty/
+
+.. _initial-release-setup:
+
+Initial Setup
+-------------
+
+.. note::
```
DanSimon, line 20: hmm these don't seem to render properly in github: https://github.com/spotify/klio/blob/bc1485a2b613156330bff6ee23863209098b2f88/RELEASING.rst

are these expected to be viewed through some other means that will render them correctly?

econchick: Oh damn, I forgot about that - originally I thought I would put them in a separate page on our docs site (which renders it fine), but then I couldn't really find a good place (I didn't want to make the contributing page suuuuper long). Thoughts on where they could be? I think I would like them on our docs site. But if we can't figure out a home, then I'll update so that it looks okay on github.

DanSimon: hmm I think it is pretty standard to have such "meta docs" in the repo's root dir.  Looking a few other repo's, they all seem to do that, for example luigi has [this](https://github.com/spotify/luigi/blob/master/RELEASE-PROCESS.rst).

So I'd suggest one of two options:

1.  just keep everything in this file and get it more compatible with github's limited rst support
2.  move most of this to a page in docs, and have this file just have a link to that

I'd _slightly_ favor option 1 just to keep everything in the same place and not over-complicate it, but I'm ok either way.

econchick: @DanSimon updated! PTAL when you're free -  I went for option 1 because I agree. Because of the limited rst support, I had to remove the `:ref:` (replaced them with anchor links), and remove the `.. note / .. attention`s and replaced them with `**note:** / **attention:**`.

DanSimon (approved):
🚀

PR #117, opened 2020-11-09
[kep] KEP-001: Purpose & Process

econchick:
Define Klio Enhancement Proposal purpose and process in the first KEP.

## Checklist for PR author(s)


docs/src/keps/kep-001.rst
```diff
@@ -0,0 +1,172 @@
...
+
+Review and Resolution
+^^^^^^^^^^^^^^^^^^^^^
+
+KEPs are discussed on their respective pull request. The possible paths of the status of KEPs are as follows:
+
+.. figure:: assets/kep_workflow.png
+    :alt: KEP Workflow
+
+    *KEP Workflow*
```
econchick, line 66: For some reason, RTDs isn't updating - I use to have a graphviz diagram here (what RTDs shows now). But then I updated it to a simple PNG. Not sure why it's not being shown. But here's what it looks like:

Screen Shot 2020-11-09 at 6 00 08 PM

docs/src/keps/kep-001.rst
```diff
@@ -0,0 +1,172 @@
...
+It is still important to have a record of this fact.
+The ``Withdrawn`` status is similar: it means that the KEP author themselves has decided that the KEP is actually a bad idea, or has accepted that a competing proposal is a better alternative.
+
+When a KEP is ``Accepted``, ``Rejected``, or ``Withdrawn``, the KEP should be updated accordingly.
+In addition to updating the status field, at the very least the ``Resolution`` header should be added with a link to the relevant PR thread.
+
+KEPs can also be ``Superseded`` by a different KEP, rendering the original obsolete.
+The ``Replaced-By`` and ``Replaces`` headers should be added to the original and new KEPs respectively.
+
+Process KEPs may also have a status of ``Active`` if they are never meant to be completed, e.g. KEP 0 (this KEP).
```
DanSimon, line 98: Does a Process KEP _have_ to go to the `Active` state?  If not, what determines when it does vs just going to `Accepted`?

Similarly, while not explicitly mentioned, it appears that Standards and Informational KEPs cannot become `Active`, only `Accepted`...is that actually the case?

Also I think this means to refer to `KEP 1` (or `KEP 001`?).

docs/src/keps/kep-001.rst
```diff
@@ -0,0 +1,172 @@
...
+All KEPs should be created as a new pull request against the ``master`` branch with the ``Draft`` status and the `"kep" GitHub tag`_.
+When a status is updated from ``Draft``, the author should update its status in the PR before merging it into ``master``.
+
+Eventually, after discussion, there may be a consensus that the KEP should be accepted – see the next section for details.
+At this point the status becomes ``Accepted``.
+
+Once a KEP has been ``Accepted``, the reference implementation must be completed.
+When the reference implementation is complete and incorporated into the main source code repository, the status will be changed to ``Final``.
+
+To allow gathering of additional design and interface feedback before committing to long term stability for a feature or API, a KEP may also be marked as "Provisional".
```
DanSimon, line 78: I think `Provisional` needs backticks instead of quotes....similar on line 82

docs/src/keps/kep-001.rst
```diff
@@ -0,0 +1,172 @@
...
+The KEP process begins with a new idea for Klio.
+It is highly recommended that a single KEP contain a single key proposal or new idea.
+Small enhancements or patches often don't need a KEP and can be injected into the Klio development workflow with a pull request to the `Klio repo`_.
+The more focused the KEP, the more successful it tends to be. If in doubt, split your KEP into several well-focused ones.
+
+Each KEP must have a champion: someone who writes the KEP using the style and format described below, shepherds the discussions in the appropriate forums, and attempts to build consensus around the idea.
+The KEP champion (a.k.a. Author) should first attempt to ascertain whether the idea is suitable for a KEP.
+This may include posting to the #klio channel in the `Spotify FOSS Slack Organization`_.
+
+The proposal should be submitted as a draft KEP via a GitHub `pull request`_ to the ``docs/src/keps`` directory with the name ``kep-<n>.rst`` where ``<n>`` is an appropriately assigned three-digit number (e.g., ``kep-000.rst``).
```
DanSimon, line 47: Who assigns the number?  Just up to the author to pick the next one?

If a PEP is withdrawn is its number recycled?  Same question for rejected but for that I assume it's not.

docs/src/keps/index.rst
```diff
@@ -0,0 +1,13 @@
+Klio Enhancement Proposals
+==========================
+
+These pages provide an overview of development decisions and priorities for Klio.
+Specifically, it contains Klio Enhancement Proposals (KEPs) – suggested changes to the ecosystem and to processes – in various stages of discussion or completion (see :doc:`KEP 0 <kep-001>`).
```
DanSimon, line 5: Is this supposed to be `KEP 001 <kep-001>`

docs/src/keps/kep-001.rst
```diff
@@ -0,0 +1,172 @@
...
+
+
+How a KEP becomes Accepted
+^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+A KEP is ``Accepted`` by consensus of all interested `core contributors`_.
+We need a concrete way to tell whether consensus has been reached.
+When you think a KEP is ready to accept, make a comment on the PR, pinging `@spotify/klio-devs`_ and asking if there are any objections within 7 days (or longer) from the comment.
+
+If all `@spotify/klio-devs`_ have subsequently approved the PR, or there have been no substantive objections by the given deadline, then the KEP can officially be marked ``Accepted`` by updating the PR and merging.
```
DanSimon, line 108: This doesn't seem to describe how a KEP can enter the `Active` state, although I imagine the steps are basically the same.

Related to my other comment, maybe it makes sense to describe two separate workflows or more clearly describe which types of PEPs can enter which states.

DanSimon (commented):
Overall this is really good.  Very clear and thorough.

docs/src/keps/kep-001.rst
econchick: A KEP does not have to go into the `Active` state. It's only in an `Active` state if it can't conceptually be completed. A Standards Track one (e.g. a feature) can be completed - therefore it goes from `Draft` -> `Accepted` -> dev process -> `Final` (the happy path I guess). A process one like this KEP-001 (or [pep-0001](https://www.python.org/dev/peps/pep-0001/), [pep-0008](https://www.python.org/dev/peps/pep-0008/)) and I'm guessing informational ones ([pep--0247](https://www.python.org/dev/peps/pep-0257/), [pep-8000](https://www.python.org/dev/peps/pep-8000/)) will always remain in an `Active` state, and just go from `Draft` -> `Active` (I guess they can be superseded too).

So yeah, Informational can become `Active`. The Standards track shouldn't really be `Active`, just go from `Draft` -> one of `Accepted`, `Provisional`, `Rejected`, or `Withdrawn`. 

I could add some clarifying language if you think that'd help?

docs/src/keps/kep-001.rst
econchick: Good q; I'll update the doc to reflect this, but basically: always increment by 1 (no real reason other than to have an explicit standard; I don't think python PEPs are strictly incremental; numpy NEPs seem to be). And no number recycling (it could be confusing, I'm guessing).

docs/src/keps/index.rst
econchick: The title of the KEP is "KEP 0" (like [PEP 1](https://www.python.org/dev/peps/pep-0001/) and [NEP 0](https://numpy.org/neps/nep-0000.html)). But it's easier to name docs with the preceding 0s because then they keep their order. I can update the language to reflect how titling should be.

docs/src/keps/kep-001.rst
econchick: Sort of related, I am going to add something like [this summary](https://www.python.org/dev/peps/) to the `index.rst`/landing page of KEPs so we can easily see get an overview of everything (I'll probably add a "status" column too).

econchick: Ah it looks like the standards track for [Python PEPs](https://www.python.org/dev/peps/#pep-status-key) are the only ones that can have the `Accepted` status.

docs/src/keps/index.rst
DanSimon: Then wouldn't this be `KEP 1` and not `KEP 0`?

econchick:
@DanSimon PTAL when you're free! I've made some updates based on your comments; I've also added a section that actually defines proposal statuses (even if it may be a bit repetitive). I also updated the `keps/index.rst` page to give a quick overview of all the available KEPs. 🙏

DanSimon (approved):
👍

PR #118, opened 2020-11-11
[lib] rewrite KlioReadFromBigQuery as reader + map transform

DanSimon:
Because beam 2.25 changed `BigQuerySource` to a function, we'll instead create `KlioReadFromBigQuery` as a custom `PTransform` that just combines beam's `ReadFromBigQuery` with a separate mapper that does the BQ row to `KlioMessage` conversion.

Beam has made quite a few breaking changes over the last few versions and this is the only approach that is compatible with both older (2.22) and newer (2.25) versions.

Tested this so far on both direct-runner and dataflow with beam 2.24 and 2.25.

This PR does _not_ remove the `<2.25` restriction on beam, but only because it leads to other google lib version conflicts.  Luckily #111 will address that (in my tests I merged that branch in to test on 2.25)

## Checklist for PR author(s)

### DIFF
--- a/lib/src/klio/transforms/io.py
+++ b/lib/src/klio/transforms/io.py
@@ -20,7 +20,6 @@ import apache_beam as beam
 
 from apache_beam.io import avroio as beam_avroio
 from apache_beam.io.gcp import bigquery as beam_bq
-from apache_beam.io.gcp import bigquery_tools as beam_bq_tools
 
 from klio_core.proto import klio_pb2
 
@@ -118,12 +117,15 @@ class KlioReadFromText(beam.io.ReadFromText, _KlioTransformMixin):
     _source_class = _KlioReadFromTextSource
 
 
-class _KlioBigQueryReader(beam_bq_tools.BigQueryReader):
-    def __init__(self, *args, klio_message_columns=None, **kwargs):
-        super(_KlioBigQueryReader, self).__init__(*args, **kwargs)
+class _KlioReadFromBigQueryMapper(object):
+    """Wrapper class to provide a ``beam.Map`` object that converts a row of a
+    ``ReadFromBigQuery`` to a properly formatted ``KlioMessage``.
+    """
+
+    def __init__(self, klio_message_columns=None):
         self.__klio_message_columns = klio_message_columns
 
-    def __generate_klio_message(self):
+    def _generate_klio_message(self):
         message = klio_pb2.KlioMessage()
         message.version = klio_pb2.Version.V2
         message.metadata.intended_recipients.anyone.SetInParent()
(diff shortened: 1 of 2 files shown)


lib/src/klio/transforms/io.py
```diff
@@ -151,47 +149,52 @@ def __iter__(self):
...
-            else:
-                data = json.dumps(row)
+    def _map_row(self, row):
+        message = self._generate_klio_message()
+        message.data.element = bytes(self._map_row_element(row), "utf-8")
+        return message.SerializeToString()
 
-            message.data.element = bytes(data, "utf-8")
-            yield message.SerializeToString()
+    def as_beam_map(self):
```
econchick, line 172: If the class itself is meant to be public (re: comment above), can you add a docstring for this one since it's a public method?

lib/src/klio/transforms/io.py
```diff
@@ -118,12 +117,11 @@ class KlioReadFromText(beam.io.ReadFromText, _KlioTransformMixin):
     _source_class = _KlioReadFromTextSource
 
 
-class _KlioBigQueryReader(beam_bq_tools.BigQueryReader):
-    def __init__(self, *args, klio_message_columns=None, **kwargs):
-        super(_KlioBigQueryReader, self).__init__(*args, **kwargs)
+class KlioReadFromBigQueryMapper(object):
```
econchick, line 120: Two things:

1. Is there a reason you made this public? I think it could be fine as semi-private. But if public is intentional, can you please add docstrings?
2. If the first answer is yes to public, then docs will have to be updated [here](https://github.com/spotify/klio/blob/master/docs/src/reference/lib/api/transforms/index.rst#docio) and [here](https://github.com/spotify/klio/blob/master/docs/src/reference/lib/api/transforms/io.rst).

econchick (approved):
Looks good to me! Just a couple of questions around docs itself, but they're not blocking.

econchick:
Followup q: you said:

> This PR does not remove the <2.25 restriction on beam, but only because it leads to other google lib version conflicts. Luckily #111 will address that (in my tests I merged that branch in to test on 2.25)

I'm a little confused (and prob a little too unfocused to think v critically right now): #111 adds the `<2.25.0` restriction to klio-exec, but you said you've tested this with 2.25....

lib/src/klio/transforms/io.py
DanSimon: ah yeah it doesn't need to be public, I'll fix that before merging.

DanSimon:
@econchick in my testing I removed the beam restriction after merging in #111.  As soon as I merge this in I'll update that branch

PR #120, opened 2020-11-18
Bumpversions cli exec devtools core lib

shireen-bean:
## Checklist for PR author(s)

### DIFF
--- a/devtools/setup.cfg
+++ b/devtools/setup.cfg
@@ -1,5 +1,5 @@
 [bumpversion]
-current_version = 0.1.0
+current_version = 0.1.1
 commit = True
 tag = True
 tag_name = devtools-{new_version}
--- a/lib/setup.cfg
+++ b/lib/setup.cfg
@@ -1,5 +1,5 @@
 [bumpversion]
-current_version = 0.2.1.post1
+current_version = 0.2.1.post2
 commit = True
 tag = True
 tag_name = lib-{new_version}
--- a/cli/setup.cfg
+++ b/cli/setup.cfg
@@ -1,5 +1,5 @@
 [bumpversion]
-current_version = 1.0.1
+current_version = 1.0.2
 tag_name = cli-{new_version}
 commit = True
 tag = True
--- a/cli/src/klio_cli/__init__.py
+++ b/cli/src/klio_cli/__init__.py
@@ -14,7 +14,7 @@
 #
 
 __author__ = "The klio developers"
-__version__ = "1.0.1"
+__version__ = "1.0.2"
 __email__ = "opensource+klio@spotify.com"
 __description__ = "Main entrypoint for Klio jobs"
 __uri__ = "https://github.com/spotify/klio"
--- a/core/setup.cfg
+++ b/core/setup.cfg
@@ -1,5 +1,5 @@
 [bumpversion]
-current_version = 0.2.0
+current_version = 0.2.1
 commit = True
 tag = True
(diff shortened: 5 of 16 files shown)


devtools/setup.cfg
```diff
@@ -1,16 +1,11 @@
 [bumpversion]
-current_version = 0.1.0
+current_version = 0.1.1
 commit = True
 tag = True
 tag_name = devtools-{new_version}
 message = [devtools] Bump version: {current_version} → {new_version}
-; for when we need to make any future dev releases
```
shireen-bean, line 7: Odd, i didnt delete this but it got deleted when i used bumpversion. this intended?

econchick: Nah that's fine - we can just look thru git history :)

econchick (changes requested):
Hey @shireen-bean - I failed to include a proper step on changelogs during the [preparing a release](https://github.com/spotify/klio/blob/master/RELEASING.rst#prepare-release) (I think that bullet point was literally a todo for me that I didn't properly write "TODO" so I missed it - lol I should fix that). But could you update all 4 packages' changelogs (they're in `docs/src/reference/<package name>/` to include what this patch release includes? They should reflect all the changes between last release and this one (e.g. not just your rcent change on moving code around, if there are more). And should be in the format of:

```rst
Changelog
========

1.0.2 (YYYY-MM-DD)
------------------

Added
~~~~~

* ...
* ...

Fixed
~~~~~

* ...
* ...

Changed
~~~~~~~

* ...

Removed
~~~~~~~

* ...

.. previous releases below
```

Only relevant sub-sections needed. And the release date should be when you're actually uploading to PyPI. ([example changelog](https://github.com/econchick/interrogate/blob/master/docs/changelog.rst)).

You can include each package's changelog updates in their respective "bumpversion" commits - but you'll have to re-git tag everything 😬 

I'll go ahead and update the releasing docs now haha.

shireen-bean:
Waiting on this to merge dsimon work with_klio_config use in klio-exec

DanSimon:
@shireen-bean I've merged my branches....

1.  we will also have to do a release of `lib`
2.  Since there's a number of significant changes going into this release, and we need to sync internal lib releases, I wonder if we should do a pre-release.  Alternatively we could bump releases as-is but only publish to test pypi until we're ready (cc @econchick, any thoughts?)

shireen-bean:
Hey @DanSimon are we ready for a release of these?
I saw that you merged your branches and i can re-make the tags and add changelongs if we are ready but wanted to make sure that associated plugins are ready for new releases.

DanSimon:
@shireen-bean yeah my changes are all in.  Since `lib` is already published I think this time around we should just go for the final releases, but maybe next time we'll try the pre-release approach (i.e. releasing `.alpha1` versions)

lib/setup.cfg
```diff
@@ -1,5 +1,5 @@
 [bumpversion]
-current_version = 0.2.1.post1
```
DanSimon, line 2: do we want to bump to `post2` or just `0.2.2`?

DanSimon (approved):
I think my question about the lib version isn't a big deal, so either way this is a 👍 for me.

lib/setup.cfg
shireen-bean: Yeah i think a `.post` release should be good here cuz there weren't any code change differences after 0.2.1, just the dependency versions which is why i chose a post

PR #121, opened 2020-11-18
[cli] Remove checks for topic/subscription in event input/output when event types are not pubsub

fallonchen:
Currently the `klio job verify` command has functionality for verifying that the job's inputs and outputs are correctly configured. This is all currently hard-coded to assume that event-I/O is pubsub and data I/O is GCS.

Going forward these checks need to be more modular so that the checks performed match the actual type of I/O.

For now we should not error out on other types of solutions.

## Checklist for PR author(s)

### DIFF
--- a/cli/src/klio_cli/commands/job/verify.py
+++ b/cli/src/klio_cli/commands/job/verify.py
@@ -329,7 +329,10 @@ class VerifyJob(object):
 
         logging.info("Verifying your event outputs...")
 
-        for output in event_outputs:
+        pubsub_event_outputs = [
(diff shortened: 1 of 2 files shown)


econchick (approved):
LGTM! It looks like you have two commits with the same message - I'm assuming it's just supposed to be one, right? Can you squash before merging?

fallonchen:
@econchick Thanks for the review! I know it looks like the same message, but they actually differ - one says input and the other output since one commit deals with event inputs and the other event outputs. Kinda curious - is this a situation where it'd be better to squash them?

PR #122, opened 2020-11-23
[DRAFT] [lib] Proposed implementation of thread concurrency management

econchick:
This is a draft of a thread concurrency management approach based on the use of a [bounded semaphore](https://docs.python.org/3/library/threading.html#threading.BoundedSemaphore). No docs or tests yet since I just want feedback on the approach before adding. CI might not yet pass either.

## Checklist for PR author(s)

### DIFF
--- a/lib/src/klio/transforms/decorators.py
+++ b/lib/src/klio/transforms/decorators.py
@@ -356,42 +356,55 @@ def __get_thread_limiter(max_thread_count, thread_limiter):
     return thread_limiter
 
 
-def _handle_klio(func_or_meth):
-    @functools.wraps(func_or_meth)
-    def method_wrapper(self, *args, **kwargs):
-        with _klio_context() as ctx:
-            setattr(self, "_klio", ctx)
+def _handle_klio(*args, max_thread_count=None, thread_limiter=None, **kwargs):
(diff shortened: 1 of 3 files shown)


lib/src/klio/transforms/decorators.py
```diff
@@ -309,42 +310,101 @@ def func_wrapper(*args, **kwargs):
...
-        if func_or_meth.__name__ == "expand":
-            return func_or_meth(self, *args, **kwargs)
-
-        wrapper = __serialize_klio_message
-        # Only the process method of a DoFn is a generator - otherwise
-        # beam can't pickle a generator
-        if __is_dofn_process_method(self, func_or_meth):
-            wrapper = __serialize_klio_message_generator
+def __get_thread_limiter(max_thread_count, thread_limiter):
+    # max_thread_count=None or True will default to CPU count
```
fallonchen, line 314: I find it a little confusing to mix boolean values into something that's supposed to be an integer - could we do something instead using enums or default constants like, 

```
#max_thread_count=None will turn off thread limits
#max_thread_count=ThreadLimits.DEFAULT will set it to default CPU count
```

econchick:
Closing as #135 is the actual implementation

PR #123, opened 2020-11-24
[lib] Add handling of yielded exceptions from functions decorated with @handle_klio

shireen-bean:
<!--- Describe your changes 

Handles when an exception is yielded in a user's process function.

Taking a look at stackdriver logs for a klio-job we saw that when an exception is yielded, the exception will hit this block of code https://github.com/spotify/klio/blob/master/lib/src/klio/transforms/decorators.py#L186:

```            for pl in payload:
                yield from __from_klio_message_generator(
                    self, kmsg, pl, incoming_item
                )```
and the exception would execute during iteration leading to a large "monster" log of the stack trace.

This PR adds Klio handling of the exception for more controlled logging of that exception.

-->

<!--- How have you tested this?
Some valid responses are:

Tested logic of executing exception within a generator externally
Tested using klio-devtools to run a job locally and yield a RuntimeError in the process function running on DirectRunner
-->

## Checklist for PR author(s)
<!-- If an item doesn't apply to your pull request, **check it anyway** to make it apparent that there's nothing left to do. -->


<!---
for more information on how to submit valuable contributions,
see https://opensource.guide/how-to-contribute/#how-to-submit-a-contribution
-->

### DIFF
--- a/lib/src/klio/transforms/decorators.py
+++ b/lib/src/klio/transforms/decorators.py
@@ -183,10 +183,30 @@ def __serialize_klio_message_generator(
 
     else:
         if isinstance(payload, types.GeneratorType):
-            for pl in payload:
-                yield from __from_klio_message_generator(
-                    self, kmsg, pl, incoming_item
+            try:
+                for pl in payload:
+                    yield from __from_klio_message_generator(
+                        self, kmsg, pl, incoming_item
+                    )
+            # This exception block will the execute
+            # if the pl item is an Exception
+            except Exception as err:
+                func_path = self.__class__.__name__ + "." + meth.__name__
+                log_msg, exc_info = __get_user_error_message(
+                    err, func_path, kmsg
                 )
+                self._klio.logger.error(log_msg, exc_info=exc_info)
+                # This will catch an exception present in the generator
+                # containing items yielded by a function/method
+                # decorated by @handle_klio.
+                # Following items in the generator will be ignored
+                # since an exception has already been detected.
+                # We won't try to serialize kmsg to bytes since
+                # something already went wrong.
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: lib/src/klio/transforms/decorators.py [hunk 1]
```diff
@@ -183,10 +183,28 @@ def __serialize_klio_message_generator(
...
         if isinstance(payload, types.GeneratorType):
-            for pl in payload:
-                yield from __from_klio_message_generator(
-                    self, kmsg, pl, incoming_item
+            try:
+                for pl in payload:
+                    yield from __from_klio_message_generator(
+                        self, kmsg, pl, incoming_item
+                    )
+            # This exception will the the pl item if it was an Exception
```
econchick, line 191: I think this comment is missing a word?

lib/src/klio/transforms/decorators.py
```diff
@@ -183,10 +183,28 @@ def __serialize_klio_message_generator(
...
+            except Exception as err:
+                func_path = self.__class__.__name__ + "." + meth.__name__
+                log_msg, exc_info = __get_user_error_message(
+                    err, func_path, kmsg
                 )
+                self._klio.logger.error(log_msg, exc_info=exc_info)
+                # This will catch an exception present in the generator
+                # containing items yielded in a job's process function.
```
econchick, line 199: Technically, a function/method that's decorated with `@handle_klio` - not necessarily a DoFn's process method. Just to be clear in the comments / not to confused future us as maintainers 😆

econchick (dismissed):
This looks good to me! Just fixup a couple of the code comments, and add a description to the PR body so that folks understand all that's going on.

PR #126, opened 2020-11-30
[docs] De-emphasize references to FnAPI

fallonchen:
Since FnAPI is still experimental (and will be for quite some time), make documentation assume that the default is to use `setup.py` for packaging, rather than FnAPI.

## Checklist for PR author(s)

- [x ] Format the pull request title like `[cli] Fixes bugs in 'klio job fake-cmd'`.

### DIFF
--- a/docs/src/faqs/index.rst
+++ b/docs/src/faqs/index.rst
@@ -27,6 +27,8 @@ FAQs
     custom_proto_def
     dags_in_klio
     migrate_from_fnapi
+    migrate_from_setup
+    setup_packaging
 
 
 General
@@ -76,3 +78,5 @@ Technical
     publish_kmsgs_from_non_klio_job
     custom_proto_def
     migrate_from_fnapi
+    migrate_from_setup
+    setup_packaging
--- a/docs/src/faqs/migrate_from_fnapi.rst
+++ b/docs/src/faqs/migrate_from_fnapi.rst
@@ -1,3 +1,5 @@
+.. _migrate-from-fnapi:
+
 How Do I Migrate from FnAPI to ``setup.py``?
 ============================================
 
--- a/docs/src/faqs/setup_packaging.rst
+++ b/docs/src/faqs/setup_packaging.rst
@@ -1,0 +1,234 @@
+What do I need to know about using ``setup.py`` to package dependencies in Klio?
+================================================================================
+
+Using ``setup.py`` file to specify dependencies to install on workers
+was the original means by which a Python Beam job developer could
+customize the environment in which their job ran.
+You can read more about it
+`here <https://beam.apache.org/documentation/sdks/python-pipeline-dependencies/>`_.
+
+To configure a Klio job to run using ``setup.py`` packaging,
+the user needs to provide the required files (see below)
+and configure the job to use them
+by filling in the ``pipeline_options.setup_file`` field in their ``klio-job.yaml`` file:
+
+
+.. code-block:: yaml
+
+    job_name: my-job
+    pipeline_options:
+        setup_file: setup.py
+        # <-- snip -->
+
+Note that this method of setup is mutually exclusive with :ref:`using the FnAPI <migrate-from-fnapi>` for packaging.
+
+Required Files
+--------------
+
+.. _setup-py:
+
+``setup.py``
+^^^^^^^^^^^^
+
+A ``setup.py`` file is needed in the **root of your job's directory**. It partly substitutes the
+need for a worker image by installing any non-Python dependencies via a child process, and by
+explicitly including non-Python files needed for a job (i.e. a model, a JSON schema, etc).
+
+
+.. tip::
+
+    The ``setup.py`` must contain the required system-level dependencies, Python dependencies, and
+    required non-Python files (i.e. ML models, JSON schemas, etc) that your job requires to run.
+
+.. collapsible:: Minimal Example ``setup.py``
+
+    The following is an example with
+    a third party Python dependency available on PyPI
+    (but no non-public Python package dependencies),
+    a non-Python file (an ML model, ``my-model.h5``),
+    and no OS-level dependencies.
+
+    .. code-block:: python
+
+        import setuptools
+
+        setuptools.setup(
+            name="my-example-job",  # required
+            version="0.0.1",  # required
+            author="klio-devs",  # optional
+            author_email="hello@example.com",  # optional
+            description="My example job using setup.py",  # optional
+            install_requires=["tensorflow"],  # optional
+            data_files=[  # required
+                (".", ["klio-job.yaml", "my-model.h5"]),
+            ],
+            include_package_data=True,  # required
+            py_modules=["run", "transforms"],  # required
+        )
+
+
+
+.. collapsible:: Example ``setup.py`` with internal & OS-level dependencies
+
+    The following is a minimal example
+    that shows a method for including internal & OS-level dependencies, in
+    addition to .
+
+    .. code-block:: python
+
+        import subprocess
+        import setuptools
+
+        from distutils.command.build import build as _build
+
+
+        # NOTE: This class is required when using custom commands (i.e.
+        # installing OS-level deps and/or deps from internal PyPI)
+        class build(_build):
+            """A build command class that will be invoked during package install.
+
+            The package built using the current setup.py will be staged and
+            later installed in the worker using `pip install package'. This
+            class will be instantiated during install for this specific scenario
+            and will trigger running the custom commands specified.
+            """
+            sub_commands = _build.sub_commands + [('CustomCommands', None)]
+
+        # `APT_COMMANDS` and `REQUIREMENTS_COMMANDS` are custom commands that
+        # will run during setup that are required for this Klio job. Each
+        # command will spawn a child process.
+        APT_COMMANDS = [
+            ["apt-get", "update"],
+            # Debian-packaged dependencies (or otherwise OS-level requirements)
+            # `--assume-yes` to avoid interactive confirmation
+            ["apt-get", "install", "--assume-yes", "libsndfile1"],
+        ]
+        REQUIREMENTS_COMMANDS = [
+            [
+                "pip3",
+                "install",
+                "--default-timeout=120",
+                # --index-url will not work on Dataflow, but `--extra-index-url` will
+                "--extra-index-url",
+                # point to your internal PyPI
+                "pypi.internal.net",
+                "--requirement",
+                "job-requirements.txt",  # Must also be included in MANIFEST.in
+            ]
+        ]
+
+
+        # NOTE: This class is required when using custom commands (i.e.
+        # installing OS-level deps and/or deps from internal PyPI)
+        class CustomCommands(setuptools.Command):
+            """A setuptools Command class able to run arbitrary commands."""
+
+            def initialize_options(self):
+                # Method must be defined when implementing custom Commands
+                pass
+
+            def finalize_options(self):
+                # Method must be defined when implementing custom Commands
+                pass
+
+            def RunCustomCommand(self, command_list):
+                # NOTE: Output from the custom commands are missing from the
+                # logs. The output of custom commands (including failures) will
+                # be logged in the worker-startup log. (BEAM-3237)
+                print("Running command: %s" % " ".join(command_list))
+                p = subprocess.Popen(
+                    command_list,
+                    stdin=subprocess.PIPE,
+                    stdout=subprocess.PIPE,
+                    stderr=subprocess.STDOUT,
+                )
+                stdout_data, _ = p.communicate()
+                print("Command output: %s" % stdout_data)
+                if p.returncode != 0:
+                    raise RuntimeError(
+                        "Command {} failed: exit code: {}".format(
+                            command_list, p.returncode
+                        )
+                    )
+
+            def run(self):
+                for command in APT_COMMANDS:
+                    self.RunCustomCommand(command)
+                for command in REQUIREMENTS_COMMANDS:
+                    self.RunCustomCommand(command)
+
+
+        # NOTE: `version` does not particularly mean anything here since we're
+        # not publishing to PyPI; it's just a required field
+        setuptools.setup(
+            name="my-example-job",  # required
+            version="0.0.1",  # required
+            author="klio-devs",  # optional
+            author_email="hello@example.com",  # optional
+            description="My example job using setup.py",  # optional
+            install_requires=[],  # optional if using the above REQUIREMENTS_COMMANDS
+            data_files=[  # required
+                # tuple(
+                #    str(dir where to install files, relative to Python modules),
+                #    list(str(non-Python filenames))
+                # )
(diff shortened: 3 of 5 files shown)


docs/src/faqs/migrate_from_fnapi.rst
```diff
@@ -1,229 +1,76 @@
...
 
-    $ klio job create --use-fnapi=false
-
-Limitations and Warnings
-------------------------
-
-* Currently, Klio in non-FnAPI mode does not yet support jobs with multiple configuration files. Support is planned.
-* ``pipeline_options.requirements_file`` configuration for `pipeline dependencies`_ **will not work** for Klio jobs. While klio will honor that configuration value for Dataflow to pick up, declaring requirements in ``setup.py`` is needed because a Klio job inherently has multiple Python files.
-* While Klio will still upload the worker image to `Google Container Registry`_ when running/deploying a job, Dataflow will *not* use the image. It is good practice to upload the worker image to ensure repeatable builds, but in the future, an option will be added to skip the upload.
-
```
econchick, line 24: I'm cool with removing this section, but I still think it's valuable information. And I'm not sure where to put it. They can probably be split up and put into different, relevant places, maybe? Any thoughts?

docs/src/faqs/migrate_from_fnapi.rst
```diff
@@ -1,229 +1,76 @@
...
-        Klio packages up your job to be installed (for unit tests, audits, and running on the direct runner), and to be uploaded to Dataflow locally on the job's worker image. Therefore, the Docker image needs to have all the required Python and non-Python files to run the job.
-
-4. **ADD** the following line to the end of the file: ``RUN pip install .``
+1. **UPDATE**  ``RUN pip install .`` to ``RUN pip install -r job-requirements.txt``
 
     .. collapsible:: Why is this needed?
 
-        We install the package for the ability to run unit tests via ``klio job test``, run audits via ``klio job audit``, and - if needed - to run the job with Direct Runner.
+       We now install dependencies directly on the worker image.
```
econchick, line 57: Looks like this `collapsible` was originally improperly indented as it looks like a block quote - can you de-dent one level? Hopefully that'll fix it and not screw up the step numbering.

Screen Shot 2020-12-01 at 3 04 49 PM

docs/src/faqs/migrate_from_fnapi.rst
```diff
@@ -1,229 +1,76 @@
...
 
 .. collapsible:: Example of Required Changes
 
     .. code-block:: diff
 
           COPY __init__.py \
-        +     setup.py \
-        +     MANIFEST.in \
         +     my-model.h5 \
-        +     klio-job.yaml \
```
econchick, line 220: So, with the current "fnapi -> setup.py" walk through as it is, it's _adding_ `my-model.h5`. I guess this particularly line shouldn't be added/shouldn't have the `+` but just there to begin with? I think it's a little confusing because you've (rightfully) deleted change # 3 under "Update: Dockerfile".

docs/src/faqs/migrate_from_fnapi.rst
```diff
@@ -1,229 +1,76 @@
...
         +     job-requirements.txt \
               run.py \
               transforms.py \
               /usr/src/app/
 
-        + RUN pip install .
+        - RUN pip install .
+        + RUN pip install -r job-requirements.txt
```
econchick, line 73: Originally, `RUN pip install .` is at the end to install the package (as defined in `setup.py`). But when using the FnAPI, it should be before the copying, since the files in `COPY` are more likely to change than `job-requirements.txt` (and therefore it's more performant to pip install deps earlier). So it actually should be something like:

```Dockerfile
...
COPY job-requirements.txt job-requirements.txt
RUN pip install -r job-requirements.txt

COPY __init__.py \
    my-model.h5 \
    run.py \
    transforms.py \
    /usr/src/app/
```

docs/src/faqs/migrate_from_fnapi.rst
```diff
@@ -233,85 +80,60 @@ The following is a collection of suggested changes to optimize Docker builds by
...
-
-    .. note::
+    **Most of these changes are incompatible with using setup.py.**
 
-        Note: Keeping ``pip install --upgrade pip setuptools`` (or similar) is still advised.
+    The following changes will break your job if you return to using ``setup.py`` to package your dependencies. If you choose to switch back, simply undo these deletions.
 
-* **DELETE** any lines creating ``/usr/src/config``, i.e. ``RUN mkdir -p /usr/src/config``.
-* **DELETE** the two lines ``ARG KLIO_CONFIG=klio-job.yaml`` and ``COPY $KLIO_CONFIG /usr/src/config/.effective-klio-job.yaml``.
+* **DELETE** lines copying ``MANIFEST.in`` and ``setup.py`` as they are no longer used.
```
econchick, line 87: Maybe warn users that if they do not follow this suggestion, and they do not actually have  `MANIFEST.in` and `setup.py` files, that building the image will break as docker will complain that it can't find them. Or make it a required change.

docs/src/faqs/migrate_from_fnapi.rst
```diff
@@ -233,85 +80,60 @@ The following is a collection of suggested changes to optimize Docker builds by
...
-* **DELETE** any lines referring to ``pip install -r job-requirements.txt`` and ``pip install klio-exec``; dependencies will be installed via the added line ``RUN pip install .``.
-
-    .. note::
+    **Most of these changes are incompatible with using setup.py.**
 
-        Note: Keeping ``pip install --upgrade pip setuptools`` (or similar) is still advised.
+    The following changes will break your job if you return to using ``setup.py`` to package your dependencies. If you choose to switch back, simply undo these deletions.
 
-* **DELETE** any lines creating ``/usr/src/config``, i.e. ``RUN mkdir -p /usr/src/config``.
-* **DELETE** the two lines ``ARG KLIO_CONFIG=klio-job.yaml`` and ``COPY $KLIO_CONFIG /usr/src/config/.effective-klio-job.yaml``.
```
econchick, line 248: Just pinging @DanSimon for when he returns - his change in #63 didn't address this - it's still in our [docker template](https://github.com/spotify/klio/blob/master/cli/src/klio_cli/commands/job/utils/templates/dockerfile.tpl#L49-L50), and this `ARG` is still populated in the [cli](https://github.com/spotify/klio/blob/master/cli/src/klio_cli/utils/docker_utils.py#L114), but I don't think this actual file is used anywhere.

docs/src/faqs/migrate_from_fnapi.rst
```diff
@@ -233,85 +80,60 @@ The following is a collection of suggested changes to optimize Docker builds by
...
+               klio-job.yaml \
                run.py \
                transforms.py \
                /usr/src/app/
 
-        -  ARG KLIO_CONFIG=klio-job.yaml
-        -  COPY $KLIO_CONFIG /usr/src/config/.effective-klio-job.yaml
-        +  RUN pip install .
+           ARG KLIO_CONFIG=klio-job.yaml
+           COPY $KLIO_CONFIG /usr/src/config/.effective-klio-job.yaml
```
econchick, line 136: You can remove these two lines.

docs/src/faqs/migrate_from_fnapi.rst
```diff
@@ -233,85 +80,60 @@ The following is a collection of suggested changes to optimize Docker builds by
...
+        -  RUN pip install .
+        +  RUN pip install -r job-requirements.txt
```
econchick, line 138: And update these lines to reflect the same suggestion that I suggested above.

Also, make sure that the example code matches the prose, and that the "Combined Example of Required & Suggested Changes" actual matches the "Example of Required Changes" plus "Example of Suggested Changes."

docs/src/faqs/migrate_from_fnapi.rst
```diff
@@ -233,85 +80,60 @@ The following is a collection of suggested changes to optimize Docker builds by
 
 .. caution::
 
-    **Most of these changes are incompatible with the FnAPI.**
-
-    The following changes will break your job if you return to using the FnAPI. If you choose to switch back to the FnAPI, simply undo these deletions.
-
-* **DELETE** any lines updating & installing Debian packages, i.e. ``apt-get update && apt-get install ...``. These commands have been moved to ``setup.py`` and will run via the added line ``RUN pip install .`` .
```
econchick, line 240: We still need to prescribe them to add  `apt-get update and apt-get install` for any system-level deps (a required change). Otherwise they won't have them by the time they install their python dependencies.

An optional/suggested change that we could provide: add `apt-get update` (but not `apt-get install`) if they do not have any system-level dependencies to install, to keep their image updated.

docs/src/faqs/migrate_from_fnapi.rst
```diff
@@ -1,229 +1,76 @@
...
-        ``setup.py`` and ``MANIFEST.in`` are needed to tell Klio and Dataflow how to build your pipeline as a Python package (i.e. what Python and non-Python files to include) since you're no longer using a Docker image as a "package" for your job.
-
-3. **DOUBLE CHECK** any non-Python files needed for the job, e.g. models, JSON schemas, etc, are copied into the working directory, ``/usr/src/app``.
-
-    .. collapsible:: Why is this needed?
-
-        Klio packages up your job to be installed (for unit tests, audits, and running on the direct runner), and to be uploaded to Dataflow locally on the job's worker image. Therefore, the Docker image needs to have all the required Python and non-Python files to run the job.
-
-4. **ADD** the following line to the end of the file: ``RUN pip install .``
+1. **UPDATE**  ``RUN pip install .`` to ``RUN pip install -r job-requirements.txt``
```
econchick, line 53: See comment below re: where to put `RUN pip install -r job-requirements.txt`. It might be easier to have two separate steps instead of one **UPDATE** step.

econchick (changes requested):

docs/src/faqs/migrate_from_fnapi.rst
fallonchen: Hmm, maybe a section under `Anatomy` called `Dependency Management`? This section would detail and explain the differences between the two options. I could also see this information going into the `FAQ` if it seems inappropriate for the `User Guide`.

fallonchen: Ugh, also on re-reading this I just realized that I screwed up a bit (not in a breaking way, but just in an accurate messaging way) with [this PR that I just merged in](https://github.com/spotify/klio/pull/127/files) since it makes it seem like you can simply provide a `requirements.txt` file.

fallonchen: Update - took a stab at creating a separate FAQ for this, called "What is needed to use setup.py to package dependencies in Klio"?

docs/src/faqs/index.rst
```diff
@@ -27,6 +27,7 @@ FAQs
     custom_proto_def
     dags_in_klio
     migrate_from_fnapi
+    setup_packaging
```
econchick, line 30: This just adds it to the left sidebar. You should also add the doc in one of the sections below (perhaps under "Technical").

docs/src/faqs/setup_packaging.rst
```diff
@@ -0,0 +1,108 @@
+What do I need to know about using ``setup.py`` to package dependencies in Klio?
+================================================================================
+
+Using ``setup.py`` or a ``requirements.txt`` file to specify dependencies to install on workers was the original means by which a Python Beam job developer could customize the environment in which their job ran. You can read more about it in the `here <https://beam.apache.org/documentation/sdks/python-pipeline-dependencies/>`_.
```
econchick, line 4: I wouldn't include "or a `requirements.txt` file" since we don't exactly support that.

econchick: Also, don't forget [semantic new lines](https://rhodesmill.org/brandon/2012/one-sentence-per-line/).

docs/src/faqs/setup_packaging.rst
```diff
@@ -0,0 +1,108 @@
...
+
+
+.. code-block:: yaml
+
+    job_name: my-job
+    pipeline_options:
+        setup_file: setup.py
+        # <-- snip -->
+
+Note that this method of setup is mutually exclusive with using the FnAPI for packaging.
```
econchick, line 16: Good opportunity to link to the other FAQ on how to migrate from setup.py to fnapi.

docs/src/faqs/setup_packaging.rst
```diff
@@ -0,0 +1,108 @@
...
+
+.. tip::
+
+    The ``setup.py`` must contain the required system-level dependencies, Python dependencies, and
+    required non-Python files (i.e. ML models, JSON schemas, etc) that your job requires to run.
+
+.. collapsible:: Minimal Example ``setup.py``
+
+    The following is an example with a non-Python file (a ML model, ``my-model.h5``), and no
+    non-public Python package dependencies or OS-level dependencies.
```
econchick, line 39: I was at first confused by "no non-public Python package" and seeing tensorflow declared there. Perhaps a slight re-word:

> The following is an example with a third party Python dependency available on PyPI (but no non-public Python package dependencies), a non-Python file (an ML model, ``my-model.h5``), and no OS-level dependencies.

docs/src/faqs/setup_packaging.rst
```diff
@@ -0,0 +1,108 @@
...
+            install_requires=["tensorflow"],  # optional
+            data_files=[  # required
+                (".", ["klio-job.yaml", "my-model.h5"]),
+            ],
+            include_package_data=True,  # required
+            py_modules=["run", "transforms"],  # required
+        )
+
+
+.. todo:: Show example with internal & OS-level dependencies
```
econchick, line 60: Do you want to address this todo now or later?

econchick (commented):
Just a few comments on the added commit.

docs/src/faqs/setup_packaging.rst
fallonchen: Added an example!

docs/src/faqs/migrate_from_fnapi.rst
```diff
@@ -1,180 +1,52 @@
-How Do I Migrate from FnAPI to ``setup.py``?
+.. _migrate-from-fnapi:
```
econchick, line 1: I just realized - this reference should probably be `_migrate-from-setup`. As well, the actual file name should be changed, particularly since that changes the URL.

Actually, I am now thinking it would be really good to keep the original `migrate_from_fnapi.rst` since folks still might want to go from fnapi to setup.py. And then have a new document, like `migrate_from_setup.rst` (or whatever you wish).

econchick (changes requested):
Sorry for taking so long to re-review. Your updates since my last review look good. I just had that one comment about keeping the original `migrate_from_fnapi.rst` file and just making a new `migrate_from_setup.rst` (or whatever you want to name it) since it'd still be useful to have "from fnapi -> setup.py".

econchick (approved):
so sorry this took me a couple of days to get to 😬  lgtm! 🚢

PR #127, opened 2020-12-02
[exec] Validate batch job packaging

fallonchen:
Since the FnAPI is not yet stable for batch jobs, warn if it is being used.

 If it is not used and neither `requirements.txt` nor `setup.py` are being used to package dependencies, exit with an error.

## Checklist for PR author(s)

- [x ] Changes are covered by unit tests (no major decrease in code coverage %) and/or integration tests.

### DIFF
--- a/exec/src/klio_exec/commands/run.py
+++ b/exec/src/klio_exec/commands/run.py
@@ -280,6 +280,13 @@ class KlioPipeline(object):
                 pipeline_opts.requirements_file
             )
 
+            if fnapi_enabled:
+                logging.warn(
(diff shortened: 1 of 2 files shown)


DanSimon (approved):
LGTM 👍 

Just one minor question: is FnApi not considered experimental for streaming?

fallonchen:
@DanSimon As I understand it, Python streaming jobs support Dataflow Runner v2/custom containers, and that Runner v2 support is nearly there for batch jobs. So technically neither are experimental, but we can expect (and warn about) instability in batch jobs, whereas it's not expected in streaming jobs.

PR #129, opened 2020-12-07
[lib] Support empty data config for helper filter transforms

econchick:
We allow users to not have to define inputs and/or outputs under `job_config.data` in their klio-job.yaml. But these helper filter transforms (`KlioFilterForce` and `KlioFilterPing`) implicitly expected data config to be there, when that shouldn’t be the case. This is a small (maybe not the most ideal) work around for that (tbh not sure what the most ideal way would be to approach this, but this was quick ¯\\_(ツ)_/¯ ).

I successfully ran jobs that have otherwise failed with issue.

## Checklist for PR author(s)

### DIFF
--- a/docs/src/reference/lib/changelog.rst
+++ b/docs/src/reference/lib/changelog.rst
@@ -1,7 +1,16 @@
 Changelog
 =========
 
-0.2.2.post (2020-12-03)
+0.2.2 (UNRELEASED)
+------------------
+
+Fixed
+*****
+
+* Allow for support of empty ``job_config.data`` values for the built-in helper filter transforms.
+
+
+0.2.1.post2 (2020-12-03)
 ------------------------
 
 Fixed
@@ -10,7 +19,7 @@ Fixed
 * Requires klio-core>=0.2.1 now that dependent changes have been released in new core version
 
 
-0.2.1.post (2020-11-30)
+0.2.1.post1 (2020-11-30)
 ------------------------
 
 Fixed
--- a/lib/src/klio/transforms/helpers.py
+++ b/lib/src/klio/transforms/helpers.py
@@ -12,6 +12,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 #
+import collections
 import logging
 
 import apache_beam as beam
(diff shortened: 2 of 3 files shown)


docs/src/reference/lib/changelog.rst
```diff
@@ -1,7 +1,16 @@
 Changelog
 =========
 
-0.2.2.post (2020-12-03)
```
econchick, line 4: This was originally not properly named - it was `0.2.1.post2` that was released

lib/src/klio/transforms/helpers.py
```diff
@@ -46,6 +51,19 @@ class KlioFilterPing(
     _helpers._KlioInputDataMixin, _helpers._KlioBaseDataExistenceCheck
 ):
     """Klio transform to tag outputs if in ping mode or not."""
+    stub_config = _StubDataConfig(ping=False, force=False)
+
+    @property
+    def _data_config(self):
+        try:
+            return super()._data_config()
```
fallonchen, line 59: Shouldn't this be `super()._data_config` (without the parentheses)?

I think this is causing this error in the integration tests:

```
    File "/usr/local/lib/python3.6/site-packages/klio/transforms/helpers.py", line 59, in _data_config
      return super()._data_config()
  TypeError: 'KlioGCSInputDataConfig' object is not callable```

File: lib/src/klio/transforms/helpers.py [hunk 3]
```diff
@@ -71,6 +89,19 @@ class KlioFilterForce(
     _helpers._KlioOutputDataMixin, _helpers._KlioBaseDataExistenceCheck
 ):
     """Klio transform to tag outputs if in force mode or not."""
+    stub_config = _StubDataConfig(ping=False, force=False)
+
+    @property
+    def _data_config(self):
+        try:
+            return super()._data_config()
```
fallonchen, line 97: This also seems like it should be `super()._data_config` (w/o parentheses)

fallonchen (changes requested):
Overall approach makes sense to me, but some of the build checks are failing.

lib/src/klio/transforms/helpers.py
econchick: Yeah I'm a derp 🤦🏻‍♀️

fallonchen (approved):
🚀

PR #131, opened 2020-12-09
[lib] Add KlioWriteToAvro

shireen-bean:
## What's New
Add ``KlioWriteToAvro`` to enable users to write an avro file.

## Decisions:
To allow this to be used as an event output write or as a generic write, we treat ``element`` as a reserved field. If the field ``element`` is present in the record to be written then that field will be set to the .
. If the field ``element`` is not present then the entire record will be cast to bytes and stuffed into the KlioMessage.data.element field.

## Testing
I have included unit tests.
One unit test to test the logic of writing to an avro file and another to test the seamless, unchaged data between reading from an avro file then writing then re-reading. (Open to getting rid of this latter test if we think it may be an overkill)

## Checklist for PR author(s)

### DIFF
--- a/lib/src/klio/transforms/io.py
+++ b/lib/src/klio/transforms/io.py
@@ -17,9 +17,9 @@ import json
 import os
 
 import apache_beam as beam
-
 from apache_beam.io import avroio as beam_avroio
 from apache_beam.io.gcp import bigquery as beam_bq
+from fastavro import parse_schema
 
 from klio_core.proto import klio_pb2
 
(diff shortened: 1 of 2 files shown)


lib/src/klio/transforms/io.py
```diff
@@ -398,3 +406,92 @@ def _get_file_pattern(self, file_pattern, location):
             file_pattern = location
 
         return file_pattern
+
+
+# note: fast avro is default for py3 on beam
+class _KlioFastAvroSink(beam_avroio._FastAvroSink):
```
DanSimon, line 412: if my struggles with BQ have been any indication, we should be careful about inheriting internal beam classes, since they seem to have no qualms with sudden breaking changes.  Could this be implemented as a custom `PTransform` that bundles a `beam.Map` with the standard `WriteToAvro`, similarly to what I did with [KlioReadFromBigQuery](https://github.com/spotify/klio/blob/master/lib/src/klio/transforms/io.py#L269)?

Don't think it's a blocker for this PR, just something we should be aware of

DanSimon (approved):
💯

PR #134, opened 2020-12-11
[exec] Remove misleading references to requirements.txt for batch packaging

fallonchen:
Since Klio jobs are inherently multi-module, a `requirements.txt` file is not sufficient to package a klio job. This PR removes the misleading check for the existence of the `requirements.txt` file.

## Checklist for PR author(s)

### DIFF
--- a/exec/src/klio_exec/commands/run.py
+++ b/exec/src/klio_exec/commands/run.py
@@ -276,10 +276,6 @@ class KlioPipeline(object):
             setup_file_exists = has_setup_file and os.path.exists(
                 pipeline_opts.setup_file
             )
-            reqs_file_exists = has_reqs_file and os.path.exists(
-                pipeline_opts.requirements_file
-            )
-
             if fnapi_enabled:
                 logging.warn(
                     "Support for batch jobs using the 'beam_fn_api' "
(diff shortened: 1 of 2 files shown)


exec/src/klio_exec/commands/run.py
```diff
@@ -287,11 +284,8 @@ def _verify_packaging(self):
                     "Use with caution."
                 )
 
-            if not any([fnapi_enabled, setup_file_exists, reqs_file_exists]):
-                logging.error(
-                    "setup.py and/or requirements file either "
-                    "unspecified or not found."
-                )
+            if not any([fnapi_enabled, setup_file_exists]):
+                logging.error("setup.py file either unspecified or not found.")
```
econchick, line 288: What do you think of adding a line, something like:

```py
if not any([fnapi_enabled, setup_file_exists]):
  if has_reqs_file:
    logging.error("Klio does not support requirements.txt file. Please use blablahbla")
  else:
    logging.error("setup.py file either unspecified or not found.")
  ...
```

exec/tests/unit/commands/test_run.py
```diff
@@ -203,7 +203,6 @@ def test_get_image_tag(image, tag, expected_image):
     assert expected_image == actual_image
 
 
-@pytest.mark.parametrize("streaming", (True, False))
```
econchick, line 206: Curious, why was this parametrization removed?

econchick (approved):
Just one question and a small suggestion that can be ignored if you wish :)

exec/src/klio_exec/commands/run.py
fallonchen: Makes sense! Added.

exec/tests/unit/commands/test_run.py
fallonchen: I went a bit too quickly, I think - I wanted this to only test cases that didn't raise and in doing so forgot that verification of the settings in `klio-job.yaml` needed to test both streaming and batch. 

I've added it back + a test case that checks for the raised exception.

exec/src/klio_exec/commands/run.py
```diff
@@ -276,23 +276,25 @@ def _verify_packaging(self):
...
+            if not any([fnapi_enabled, setup_file_exists]):
+                if has_reqs_file:
+                    logging.warn(
+                        "Klio jobs are multi-module. "
+                        "Thus, a setup.py file is required "
+                        "in addition to a requirements.txt file"
+                    )
+                else:
+                    logging.error(
+                        "setup.py file either unspecified " "or not found."
```
econchick, line 295: Nit: you have an extraneous `" "` in the middle there

econchick (dismissed):

exec/src/klio_exec/commands/run.py
fallonchen: Ah! Nice catch, fixed :)

PR #135, opened 2020-12-14
[lib, docs] Add thread management to Klio-ified transforms

econchick:
This PR implements [KEP 2](https://docs.klio.io/en/latest/keps/kep-002.html) and is based off of #122. 

I've included unit tests and have successfully run jobs with the added code.

## Checklist for PR author(s)

### DIFF
--- a/docs/src/userguide/pipeline/utilities.rst
+++ b/docs/src/userguide/pipeline/utilities.rst
@@ -1,7 +1,7 @@
 Utilities
 =========
 
-Klio offers decorators that help Klio-ify transforms of a pipeline.
+Klio offers decorators and other utilities that help Klio-ify transforms of a pipeline.
 
 
 .. _serialization-klio-message:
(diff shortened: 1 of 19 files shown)


integration/audio-spectrograms/run.py
```diff
@@ -37,6 +37,9 @@
 for logger in loggers_to_mute:
     logging.getLogger(logger).setLevel(logging.ERROR)
 logging.getLogger("klio").setLevel(logging.DEBUG)
+# The concurrency logs may be different for every machine, so let's
+# just turn them off
+logging.getLogger("klio.concurrency").setLevel(logging.ERROR)
```
econchick, line 42: FYI I didn't apply this line to _all_ integration tests. The ones I didn't apply them to don't actually look at the job output logs at all.

lib/tests/unit/transforms/test_helpers.py
```diff
@@ -404,7 +408,8 @@ def test_update_klio_log(mocker, monkeypatch, caplog, mock_config):
         assert False, "Expected debug audit log not found"
 
 
-def test_trigger_upstream_job(mock_config, mocker):
+@pytest.mark.skipif(IS_PY36, reason="This test fails to pickle on 3.6")
```
econchick, line 411: So this was really annoying. This test fails _only_ on Python 3.6 (regardless of beam version), but it passes with the `-s / --capture=no` flag. It also isn't reproducible when running an actual job (direct runner or dataflow). The error I get is `TypeError: cannot serialize 'EncodedFile' object`. The `EncodedFile` object looks like it comes from [pytest](https://github.com/pytest-dev/pytest/blob/master/src/_pytest/capture.py#L178). But I'm not sure how to actually address it. I wanted to get this patch out the door so I marked it as to skip for now when in a py36 environment. But I'll make a ticket to remember to investigate further.

integration/batch-modular-default/run.py
```diff
@@ -31,6 +31,9 @@
 )
 for logger in loggers_to_mute:
     logging.getLogger(logger).setLevel(logging.ERROR)
+# The concurrency logs may be different for every machine, so let's
+# just turn them off
+logging.getLogger("klio.concurrency").setLevel(logging.ERROR)
```
shireen-bean, line 36: Just wondering if there is a reason `"klio.concurrency"` is turned off explicitly rather than added to `loggers_to_mute`.

econchick: ah...tbh I didn't even think to put them in `loggers_to_mute` 😆 good idea

lib/src/klio/utils/_thread_limiter.py
```diff
@@ -0,0 +1,157 @@
...
+        thread_limiter = ThreadLimiter(max_thread_count=limit_func)
+
+        # Turn off thread limiting
+        thread_limiter = ThreadLimiter(max_thread_count=ThreadLimit.NONE)
+
+    Args:
+        max_thread_count (int, callable, ThreadLimit): number of threads
+            to make available to the limiter, or a :func:`callable` that
+            returns an ``int``. Set to :attr:`ThreadLimit.NONE` for no
+            thread limits.
```
fallonchen, line 96: Should we mention that if this is < 1, a `ValueError` will be raised? If people provide a lambda and implement it incorrectly, this could happen.

lib/src/klio/utils/_thread_limiter.py
```diff
@@ -0,0 +1,157 @@
...
+        thread_limiter = ThreadLimiter()
+
+        # Then use it as a context manager:
+        with thread_limiter:
+            ...
+
+        # Or explicitly call acquire and release:
+        thread_limiter.acquire()
+        ...
+        thread_limiter.release()
```
fallonchen, line 77: If we're not explicitly calling out the fact that we're using a `BoundedSemaphore` under the hood, it could be good to say that when the `thread_limiter` is released too many times, a `ValueError` will be raised.

lib/src/klio/utils/_thread_limiter.py
```diff
@@ -0,0 +1,157 @@
...
+        multi-threaded environment.
+
+        If no semaphores are available, the method will block until one
+        is released via :func:`ThreadLimiter.release`.
+        """
+        self._log("Blocked – waiting on semaphore for an available thread")
+        self._semaphore.acquire()
+
+    def release(self):
+        """Release a semaphore (a thread)."""
```
fallonchen, line 145: If we're not explicitly calling out the fact that we're using a BoundedSemaphore under the hood, it could be good to say that when the thread_limiter is released too many times, a ValueError will be raised.

docs/src/userguide/pipeline/utilities.rst
```diff
@@ -350,3 +350,53 @@ effect.
...
+.. _klio-concurrency-mgmt:
+
+Concurrency Management
+----------------------
+
+Currently, Apache Beam does not have a mechanism to limit the number of threads that a pipeline uses to execute its transforms.
+Klio provides a way for users to limit the number of threads used for a specific transform or section of code, with defaults set to the number of CPUs of the worker machine.
+See :doc:`../../../keps/kep-002` for more background information.
+
+There are two mechanisms to manages threads in a Klio pipeline:
```
fallonchen, line 364: s/manages/manage

fallonchen (dismissed):
🎉 

Just one minor typo in the docs, feel free to merge after that's addressed! 

I also left a couple questions about making the error states more visible for people using `ThreadLimiter` objects directly, but they're not blockers to merging.

shireen-bean (approved):
Looks like fallon approved and you handled all her comments in addition to mine (ty!)

PR #139, opened 2020-12-16
[core] fix preprocessing of config io fields

DanSimon:
Looks like we weren't actually applying the transform on raw config that converted the lists of I/O to nested dicts.  Luckily this had no real effect except on correctly handling overrides.

## Checklist for PR author(s)

### DIFF
--- a/core/src/klio_core/config/core.py
+++ b/core/src/klio_core/config/core.py
@@ -18,6 +18,7 @@ import logging
 import attr
 
 from klio_core.config import _io as io
+from klio_core.config import _preprocessing as preprocessing
 from klio_core.config import _utils as utils
 
 logger = logging.getLogger("klio")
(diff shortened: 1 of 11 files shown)


core/src/klio_core/config/core.py
```diff
@@ -83,6 +84,24 @@ class KlioConfig(BaseKlioConfig):
         job_config (KlioJobConfig): Job-related configuration.
     """
 
+    def __init__(
+        self,
+        raw_config_data,
+        raw_overrides=[],
+        raw_templates=[],
```
econchick, line 91: For these two args, I'd suggest setting them to `None`, then in the actual function, set them to an empty list if they're equal to `None`. Setting defaults to something mutable [can cause unexpected behavior](https://stackoverflow.com/questions/1132941/least-astonishment-and-the-mutable-default-argument), so unless there's a clear reason why we can't avoid it - even if it's not possible to encounter this unexpected behavior - I'd prefer to avoid it entirely.

core/src/klio_core/config/core.py
```diff
@@ -216,7 +235,7 @@ def _create_config_objects(self, configs, io_type, io_direction):
             ]
         )
         objs = []
-        for config in configs:
+        for (name, config) in configs.items():
```
econchick, line 238: There's no need to make this into a tuple:

```py
for name, config in configs.items():
```

docs/src/reference/core/changelog.rst
```diff
@@ -1,6 +1,14 @@
 Changelog
 =========
 
+0.2.2 (UNRELEASED)
+------------------
+
+Fixed
+*****
+
+* fix preprocessing of config io fields
```
econchick, line 10: Mind giving more of an explanation? It can still be a single sentence - just something that makes more sense for the user (i.e. include what was the problem)

econchick (changes requested):
Looks good to me - just a couple of small nits

PR #141, opened 2021-01-04
[lib] Bugfix: klio serialization in helpers & for non-klio messages.

econchick:
When working on my Klio pet, I noticed two things wrong with Klio serialization:

1. When working with non-Klio messages, Klio will load the whole non-Klio message into `kmsg.data.element` (as expected). But it did not set the message version or the recipients, both of which are expected in other parts of the code. The first commit addresses that.
2. Some of the helper transforms take in a raw pubsub message and need to serialize it into a Klio message. They were not providing all the required function arguments when serializing to Klio message, and the second commit addresses that.

* Updated unit tests
* Successfully ran my Klio pet job with these changes

## Checklist for PR author(s)

### DIFF
--- a/lib/src/klio/message/serializer.py
+++ b/lib/src/klio/message/serializer.py
@@ -92,6 +92,10 @@ def to_klio_message(incoming_message, kconfig=None, logger=None):
             # We are assuming that we have been given "raw" data that is not in
             # the form of a serialized KlioMessage.
             parsed_message.data.element = incoming_message
+            # default to set recipients to anyone - can't know who the
+            # appropriate recipient is when it's not a real klio msg
+            parsed_message.metadata.intended_recipients.anyone.SetInParent()
+            parsed_message.version = klio_pb2.Version.V2
         else:
             logger.error(
                 "Can not parse incoming message. To support non-Klio "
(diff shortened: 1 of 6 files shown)


lib/src/klio/message/serializer.py
```diff
@@ -92,6 +92,10 @@ def to_klio_message(incoming_message, kconfig=None, logger=None):
             # We are assuming that we have been given "raw" data that is not in
             # the form of a serialized KlioMessage.
             parsed_message.data.element = incoming_message
+            # default to set recipients to anyone - can't know who the
+            # appropriate recipient is when it's not a real klio msg
+            parsed_message.metadata.intended_recipients.anyone.SetInParent()
+            parsed_message.version = 2
```
shireen-bean, line 98: Would we want to set this to `klio_pb2.Version.V2` similar to how we do in the KlioReadFromXXX transforms?

econchick: ah yeah sure (fwiw both work, but I agree, I like this explicit-ness)

econchick:
@shireen-bean I addressed your comment! PTAL (and anyone else still reviewing) when you have a moment 🙂

shireen-bean (approved):

PR #144, opened 2021-01-08
[core] fixing config as_dict not including custom keys on dataflow

DanSimon:
This fixes a bug where calling `klio_config.as_dict()` in a transform would not include user-defined config keys only when executed in dataflow.

Basically `USER_ATTRIBS` was defined as a class-level attribute, which means it would not get pickled with other class attributes when sent to dataflow.  This explains that while the keys didn't appear in the generated dict, they did exist on dataflow through direct access.

## Checklist for PR author(s)

### DIFF
--- a/core/setup.py
+++ b/core/setup.py
@@ -170,6 +170,7 @@ EXTRAS_REQUIRE = {
         "pytest>=4.3.0",  # 4.3.0 dropped last use of `convert`
         "pytest-cov",
         "pytest-mock",
+        "dill",
     ],
 }
 EXTRAS_REQUIRE["dev"] = (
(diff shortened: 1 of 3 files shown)


econchick (commented):
ah ha! good catch. Is it easy enough to add a test for this? Maybe a simple unit test that pickles an instance, then unpickles, and asserts expected attributes?

DanSimon:
@econchick I actually tried to write a test but removed it because it was passing without the bug fix.  hmm actually thinking now as I type...I can try to write one where I clear out all class-level attributes after pickling, to attempt to replicate what happens when a dataflow worker unpickes...

econchick:
@DanSimon wouldn't it defeat the purpose of the test to clear out class-level attributes after unpickling? that seems like making the unit test behave the way you want it to. Maybe using `dill` instead of stdlib's `pickle` provides different results (since beam uses `dill`)?

econchick:
nonetheless - feel free to merge this w/o a test right now to get it released sooner

PR #147, opened 2021-01-12
[exec,lib] partial revert of using main_session for KlioConfig

DanSimon:
This is a partial revert for f58caf7224807793082b922d368acff03ba76978.

In some cases for unknown reasons, `RunConfig.get()` will begin raising an exception that the config is no longer set.  This should not be happening, and in some instances the problem doesn't show up until a streaming job has been running for hours, but there is likely some underlying issue in Beam/Dataflow that is leading the main session variable to be deleted.

This PR is not intended to be a complete fix, but rather a mitigation that reverts back to `klioexec` getting its config via the `effective-klio-job.yaml` that should be included with the packaged code.

I've left the `klio-cli` changes untouched, meaning it will continue to write a temp file and bind it for use by `klioexec`.  `klioexec` will also still set the `KlioConfig` to the main session variable with `RunConfig.set()`, even though for now it won't be used.

## Checklist for PR author(s)

### DIFF
--- a/exec/src/klio_exec/cli.py
+++ b/exec/src/klio_exec/cli.py
@@ -57,9 +57,11 @@ def _get_config(config_path):
         raise SystemExit(1)
 
 
-# TODO : remove this when the call in spotify-klio is removed
-def _compare_runtime_to_buildtime_config(runtime_config_path):
-    return True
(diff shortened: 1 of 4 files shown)


fallonchen (approved):
👍

PR #158, opened 2021-01-27
[docs] Add prose docs on templating and overriding config

econchick:
As the PR title says, I've added some prose documentation that is meant to complement the CLI's reference/API docs for templating and overriding values in a job's configuration.

## Checklist for PR author(s)

### DIFF
--- a/docs/src/userguide/config/index.rst
+++ b/docs/src/userguide/config/index.rst
@@ -144,6 +144,121 @@ access to configuration is needed for a transform's logic, then use the provided
             ...
 
 
+Templating and Overriding a Job's Configuration
+-----------------------------------------------
+
+Klio supports templating ``klio-job.yaml`` as well as overriding the configuration's values via the CLI.
+
+Templating Keys and Values
+^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+To template any key or value in ``klio-job.yaml``, declare it with ``${...}``, for example:
+
+.. code-block:: yaml
+
+  # klio-job.yaml snippet
+  job_name: my-job
+  job_config:
+    a_key: ${my_template_value}
+    ${a_template_key}: some_static_value
+
+
+Then, use the ``--template`` flag (supported on most ``klio job`` subcommands – see :doc:`CLI documentation </reference/cli/api/cli/job>` for which commands supports the flag) to provide the content for the templated variables:
+
+.. code-block:: sh
+
+  # command accepts multiple `--template` flags
+  $ klio job run \
+    --template my_template_value=foo \
+    --template a_template_key=some_variable_key
+
+
+Within your Klio job, you'll have access to these configuration values just like normal:
+
+.. code-block:: python
+
+    # in transforms.py
+
+    import apache_beam as beam
+    from klio.transforms import decorators
+
+    class MyTransform(beam.DoFn):
+        @decorators.handle_klio
+        def process(self, item):
+            my_templated_value = self._klio.config.job_config.a_key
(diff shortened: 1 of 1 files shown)


docs/src/userguide/config/index.rst
```diff
@@ -144,6 +144,103 @@ access to configuration is needed for a transform's logic, then use the provided
...
+    ${a_template_key}: some_static_value
+
+
+Then, use the ``--template`` flag (supported on most ``klio job`` subcommands – see :doc:`CLI documentation </reference/cli/api/cli/job>` for which commands supports the flag) to provide the content for the templated variables:
+
+.. code-block:: sh
+
+  # command accepts multiple `--template` flags
+  $ klio job run \
+    --template my_template_value=foo \
```
DanSimon, line 172: Maybe someone else can verify, but I've only been able to get this to work when it's formatted as `--template=my_template_value=foo`, using `=` instead of a space.  Same for overrides or apparently any multi-value option.

docs/src/userguide/config/index.rst
```diff
@@ -144,6 +144,103 @@ access to configuration is needed for a transform's logic, then use the provided
...
+.. code-block:: sh
+
+  # command accepts multiple `--override` flags
+  $ klio job run \
+    --override job_config.a_key=a_new_value \
+    --override job_config.b_key=b_new_value
+
+.. attention::
+
+  Be sure to provide the full path to the desired key to change.
```
DanSimon, line 222: I think it might be good to also call out that if users want to override a value for an input/output, the easiest way to is to give it a name, otherwise Klio ends up auto-generating one (which you can still use in overrides, but isn't very obvious).

Maybe something like:
```
If you want to override a value in a configured data/event input or output, be sure to 
specify a name in its config, and then use the name in the path of the override

.. code-block:: yaml

  job_config:
    events:
      inputs:
        - type: pubsub
          name: my_input
          topic: my/pubsub/topic
          subscription: my/pubsub/subscription

.. code-block:: sh

  klio job run \
    --override=job_config.events.inputs.my_input.topic=/my/other/pubsub/topic
```

DanSimon (commented):
couple minor comments, looks pretty good overall 💯

docs/src/userguide/config/index.rst
econchick: I actually did verify that this worked. In `klio_cli.cli.py`, I modified the `run` function like so:

```py
@job.command("run", help="Run a klio job.")
@core_options.image_tag(default=None, show_default="``git-sha[dirty?]``")
@options.runtime
@core_utils.with_klio_config
def run_job(klio_config, config_meta, **kwargs):
    import json
    print(json.dumps(klio_config.as_dict()))
    return
```

And then just ran
```sh
klio job run \
  --template my_template_value=foo \
  --template a_template_key=some_variable_key \
  | jq .job_config
```

and all looked good to me. I just ran it again with `--template=my_template_value=foo` and it's the same behavior.

DanSimon: yep I just went back and double checked and it worked for me as well.  Not sure what I was doing when I initially had issues 🤷

docs/src/userguide/config/index.rst
econchick: ahh good catch - thank you!

econchick:
@DanSimon thanks for the review! I added the bit about overriding IOs. Let me know if you think I'm missing something re `--template foo=bar` vs `--template=foo=bar` since they work the same for me.

PR #161, opened 2021-02-01
[cli,core,exec,lib] include runtime config in setup.py packaging

DanSimon:
_Status - ready for review_

This is an approach to support transporting runtime config in jobs that use setup.py.  This will not work with any jobs using FnApi.

* `materialized_config.yaml` has been replaced with `klio-job-run-effective.yaml` and is used by both `klio-cli` and `klioexec`
  * `klio-cli` no longer mounts this file into the docker container, and instead just writes it to the `job_dir`.  It still invokes `klioexec` with the `--config-file` flag, which now points it to `/usr/local/app/klio-job-run-effective.yaml`.
  * `klioexec` will write its config to the same path before starting the Beam pipeline, so that it can be included in the python distribution package.
* `klio-lib` will look for `klio-job-run-effective.yaml` before looking for `.effective-klio-job.yaml`.  When using FnApi, `.effective-klio-job.yaml` will continue to be supported (since existing jobs have the `COPY` setup in their `Dockerfile`)

This allows us to handle both situations:
* When `klioexec` is run via `klio-cli`, `klioexec`'s write of the config will essentially be a no-op, since it'll write exactly what it read in.
* When `klioexec` is run directly, potentially with cli overrides/templates, it'll include them when it writes the config.

This requires the following changes in the job itself:

1. `klio-job-run-effective.yaml` must added to `data_files` in `setup.py`
2.  The following lines must be added to the job's `Dockerfile` **before** the `pip install .` line (and can replace similar lines for `.effective-klio-job.yaml`).

```
ARG KLIO_CONFIG=klio-job.yaml
COPY $KLIO_CONFIG klio-job-run-effective.yaml
```

## Checklist for PR author(s)


econchick (commented):
In the future, I would like to see a separate commit for just moving code around (i.e. the new `write_to_file` method in `klio-core` moved from `klio-cli`). It would be a lot easier to review.

Overall, I think this would work, but it's quite confusing with all of our config files 😬 and it may be valuable to take a small step back to re-evaluate the config file variations.

lib/src/klio/transforms/core.py
```diff
@@ -33,21 +34,38 @@ class RunConfig(object):
...
+        if not klio_job_file:
             # use iterator so we don't waste time searching everywhere upfront
             files = glob.iglob("/usr/**/klio-job.yaml", recursive=True)
             for f in files:
                 klio_job_file = f
                 # only grab the first one
                 break
+
+        if not klio_job_file:
+            klio_job_file = "/usr/src/config/.effective-klio-job.yaml"
```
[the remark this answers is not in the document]
DanSimon: I have to go back and retrace my steps.  I initially did try to consolidate all these different config files (which is why I moved the `materialized_config` path into `klio_core`) but ran into issues, but thinking back they may have been unrelated.  I do fully agree though that having both `materialized_config` and `run_config` is too much.

Also, `.effective-klio-job.yaml` is created in the job's docker image, not by Klio itself, so it's going to be difficult for us to "move" it.  We have to be careful about maintaining backwards compatibility.

exec/src/klio_exec/cli.py
```diff
@@ -58,10 +58,7 @@ def _get_config(config_path):
 
 
 def _compare_runtime_to_buildtime_config(klio_config):
-    buildtime_config_path = "/usr/src/config/.effective-klio-job.yaml"
-    buildtime_config = config.KlioConfig(_get_config(buildtime_config_path))
-
-    return buildtime_config.as_dict() == klio_config.as_dict()
+    return True
```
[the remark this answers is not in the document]
DanSimon: internal code calls this so we need to keep it around for now

exec/src/klio_exec/commands/run.py
```diff
@@ -260,11 +270,25 @@ def _get_run_callable(self):
             )
             raise SystemExit(1)
 
+    def _write_run_effective_config(self):
+        # this method assumes setup.py is being used!
+        path = config_core.RUN_EFFECTIVE_CONFIG_PATH
+        logging.debug(
+            "dumping runtime configuration to {}"
```
econchick, line 277: Small nitpick - could it be "Writing" instead of "dumping"? Either way, I suggest titlecasing the first word as well as `Docker`. 

Also, maybe this is just me, but is a little misleading to say "in job's Docker image", since it's not in _the_ image? just the container? Maybe, `"Writing runtime configuration to {} in the job's running Docker container."`?

core/src/klio_core/config/core.py
```diff
@@ -109,6 +111,31 @@ def as_dict(self):
...
+                Dumper=utils.IndentListDumper,
+                default_flow_style=False,
+                sort_keys=False,
+            )
+
+        if isinstance(path_or_stream, str):
+            dirname = os.path.dirname(path_or_stream)
+            if not os.path.exists(dirname):
+                os.makedirs(dirname)
+            with open(path_or_stream, "w+") as stream:
```
econchick, line 142: Curious, why the `+`?

lib/src/klio/transforms/core.py
```diff
@@ -35,19 +36,29 @@ class RunConfig(object):
...
+
+        logger = logging.getLogger("klio")
+        logger.debug(f"loading config file from {klio_job_file}")
```
econchick, line 60: Same nitpick her w/ titlecasing 😅

econchick (changes requested):
Just to be transparent to lurkers since we talked privately on slack:

A few additional changes are needed:

* a commit that changes [this line](https://github.com/spotify/klio/blob/master/cli/src/klio_cli/commands/job/utils/templates/setup.py.tpl#L129) in the job template 
* some way to deal with compatibility; for instance, an error saying something like “since klio v 1.2.3 setup.py should now include blablablah”
* update FAQs that touch on this ([one](https://docs.klio.io/en/latest/faqs/migrate_from_setup.html), [two](https://docs.klio.io/en/latest/faqs/setup_packaging.html), [three](https://docs.klio.io/en/latest/faqs/migrate_from_fnapi.html)), even if it’s just the code examples itself and not necessarily prose
* respective changelog entries

cli/src/klio_cli/commands/job/utils/templates/MANIFEST.in.tpl
```diff
@@ -1 +1,2 @@
 include job-requirements.txt
+include klio-job-run-effective.yaml
```
shireen-bean, line 2: Are these template changes changes that we should make in our cookie cutters too? (i can definitely help with that when this is close to merging if so)

econchick, line 2: Is this indeed needed? (referring to our own [docs](https://docs.klio.io/en/latest/faqs/migrate_from_fnapi.html#new-manifest-in) - not including `yaml` files weren't needed before; just non-Python files needed to install the package)

DanSimon: ugh, apparently it isn't.  I tested this a bunch but now I think there was an issue with my test setup.  I'll remove.

cli/src/klio_cli/commands/job/utils/templates/dockerfile.tpl
```diff
@@ -46,7 +46,11 @@ COPY __init__.py \
...
+COPY $KLIO_CONFIG klio-job-run-effective.yaml
+
 RUN pip install .
-{% endif -%}
+{%- else -%}
 ARG KLIO_CONFIG=klio-job.yaml
-COPY $KLIO_CONFIG /usr/src/config/.effective-klio-job.yaml
\ No newline at end of file
+COPY $KLIO_CONFIG /usr/src/config/.effective-klio-job.yaml
+{% endif -%}
```
econchick, line 56: I'm curious why you've added `else` here. Both if/else branches have the same `ARG`/`COPY` lines - I feel like this should have just changed the name in the `COPY` line; or am I missing something?

econchick (dismissed):
Just one small question, but it's a non-blocker.

🎉 🚀

cli/src/klio_cli/commands/job/utils/templates/dockerfile.tpl
DanSimon: ah yeah it shouldn't have the `ARG` line

PR #162, opened 2021-02-01
[docs] Add prose docs on multiple event inputs

econchick:
I noticed we didn't have any prose docs on multiple event inputs so I quickly drafted some up 😄 

## Checklist for PR author(s)

### DIFF
--- a/docs/src/userguide/config/job_config.rst
+++ b/docs/src/userguide/config/job_config.rst
@@ -63,7 +63,7 @@ Example:
 A list of input configuration that will be will used to determine when and how to do work.
 
 If more than one input is configured, please familiarized yourself with
-:doc:`how multiple configured inputs <../pipeline/multiple_inputs>` are handled in Klio.
+:ref:`how multiple configured inputs <multiple-event-inputs>` are handled in Klio.
 
 
 .. option:: job_config.events.inputs[].type STR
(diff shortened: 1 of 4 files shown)


docs/src/userguide/io/index.rst
```diff
@@ -88,8 +88,66 @@ object, and it will be the responsibility of your own code to setup the input.
...
+  job_config:
+    events:
+      inputs:
+        - type: file
+          location: ./batch_track_ids_1.txt
+        - type: file
+          location: ./batch_track_ids_2.txt
+    # <-- snip -->
+
+Klio will automatically read from the configured inputs.
```
shireen-bean, line 118: Does it read them in order or simultaneously?

shireen-bean: oh looks like this is covered below!

econchick: It is probably a good idea to clarify it up here even if it's covered below, just to be super clear for the reader. Good catch!

shireen-bean (approved):
This sounds clear to me!

Tangential to this - I wonder if it would make sense to have a section to organize the different run.py arguments depending on configuration somewhere. Ex: the first argument into `def run(x, y)` could be a single input p_coll, it could be multiple `pcolls` with multiple inputs, and in batch it could be a pipeline object correct? Those things are all documented in our docs in separate logical places but im wondering if it would make sense in in the overview or anotomy::run.py section too.

econchick:
@shireen-bean I think that's a good idea. I'll take that into account for when I continue my docs re-organization :) (also, it could be a pipeline object in either batch or streaming; it's when the user turns off klio's automatic handling of event inputs)

PR #169, opened 2021-02-16
Prepare candidate releases for cli, core, exec, lib

DanSimon:
## Checklist for PR author(s)


docs/src/reference/cli/changelog.rst
```diff
@@ -1,7 +1,7 @@
 CLI Changelog
 =============
 
-1.0.6 (UNRELEASED)
+1.0.6 (1.0.6rc1)
```
econchick, line 4: These should be dates - sorry if I misspoke during our convo. Something like: `1.0.6rc1 (2021-02-16)`

econchick (changes requested):
Just one thing about the changelogs. Also I don't yet see the git tags pushed.

docs/src/reference/cli/changelog.rst
DanSimon: ah ok that makes more sense.  One question, if we have to publish an `rc2`, or when we do the final release, will we be editing this line or will we add a new entry to the changelog?

econchick: I would say new entry/section

DanSimon: ok, and would we copy the lines out of the `rc` releases or would the final release entry just be empty?

econchick: ah v good question! Thinking about this a little bit more, I think we should change the header rather than the add a new entry section (and of course add any bullets of user facing changes between last rc and this rc/GA release).

econchick (approved):

PR #177, opened 2021-03-09
[exec] Fix: use --config-file override in test command

denis-angilella:
klioexec test --config-file option was ignored
Unit test added

## Checklist for PR author(s)

- [X] Format the pull request title like `[cli] Fixes bugs in 'klio job fake-cmd'`.
- [X] Changes are covered by unit tests (no major decrease in code coverage %) and/or integration tests.
- [X] already documented but not working / Document any relevant additions/changes in the appropriate spot in `docs/src`.
- [X] For any change that affects users, update the package's changelog in ``docs/src/reference/<package>/changelog.rst``.

PR #180, opened 2021-03-18
[cli] Enable job test command to use the --config-file option

fallonchen:
`klio job test` currently takes in the `--config-file` option but does nothing with it, since the `requires_config_file` flag is set to `False`. 

The config file option was also ignored in `klio-exec`, but that was fixed via [PR 177](https://github.com/spotify/klio/pull/177).

This PR sets enables the use of the `--config-file` option by doing 2 things:

1. Setting `TestPipeline().requires_config_file` to `True`. 
2. This PR also adjusts the way `TestPipeline` handles command generation so that the `--config-file` option is appended after the `test` command, but before pytest arguments are added. 

Before this change, the generated command would be: 

```klio job test -- py_arg1 pyarg2 ... pyargN --config-file config.yaml```

This causes the `--config-file` option to be ignored. 

After this change, the generated command would be:

```klio job test --config-file config.yaml -- py_arg1 pyarg 2 .. pyargN```

## Checklist for PR author(s)

### DIFF
--- a/cli/src/klio_cli/commands/job/test.py
+++ b/cli/src/klio_cli/commands/job/test.py
@@ -21,12 +21,23 @@ class TestPipeline(base.BaseDockerizedPipeline):
 
     def __init__(self, job_dir, klio_config, docker_runtime_config):
         super().__init__(job_dir, klio_config, docker_runtime_config)
-        self.requires_config_file = False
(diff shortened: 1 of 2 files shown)


DanSimon (approved):
👍

PR #184, opened 2021-03-29
[lib, docs] Add default metrics for batch IO transforms, and retry & timeout decorator

econchick:
This is part 1 of adding default metrics. Part 2 seems to be a bit more hairy (more info below).

This PR adds code to emit metrics for the following transform classes, with the name of the metric in parens (the name shows up in the dataflow UI):

* io transforms:
  * `KlioReadFromText`  (`kmsg-read`)
  * `KlioReadFromBigQuery` (`kmsg-read`)
  * `KlioReadFromAvro` (`kmsg-read`)
  * `KlioWriteToText` (`kmsg-write`)
  * `KlioWriteToBigQuery` (`kmsg-write`)
  * `KlioWriteToAvro` (`kmsg-write`)
* helper transforms:
  * `KlioGcsCheckInputExists` (`kmsg-data-found-input`, `kmsg-data-not-found-input`)
  * `KlioGcsCheckOutputExists` (`kmsg-data-found-output`, `kmsg-data-not-found-output`)
  * `KlioFilterPing` (`kmsg-process-ping`, `kmsg-skip-ping`)
  * `KlioFilterForce` (`kmsg-process-force`, `kmsg-skip-force`)
  * `KlioWriteToEventOutput` (`kmsg-output`) (I just realized, maybe this should be `kmsg-write`?)
  * `KlioDrop` (`kmsg-drop`)
  * `KlioCheckRecipients` (`kmsg-drop-not-recipient`)
  * `KlioDebugMessage` (`kmsg-debug`)
  * `KlioTriggerUpstream` (`kmsg-trigger-upstream`)
* decorators:
  * `@retry` (`kmsg-retry-attempt`, `kmsg-drop-retry-error`)
  * `@timeout` (`kmsg-drop-timed-out`)

You might have noticed, I tried to name the metrics in a way that makes sense. Something like `kmsg-<what>(-<specifier>)`. But please let me know if things could be named better!

This PR also includes a new helper transform, `KlioMessageCounter`. In a lot of cases, it was easier to have a separate transform to then use in a composite. And I figured, why not make it a helper transform for folks.

What's visible in the Dataflow UI:

image

_(note: the `kmsg-drop-error` and the pubsub read/write counters are not a part of this PR but will be for the next one)_ 

These metrics also show up in Stackdriver monitoring. Below is a screenshot of these counters being available to select in [Metrics Explorer](https://console.cloud.google.com/monitoring/metrics-explorer) when building a dashboard:

image

In part two, I'm hoping to implement the following metrics:

* read/write counters for PubSub 
* `kmsg-drop-error` for any error caught in `@handle_klio`
* `kmsg-timer` to measure the time it takes to process a message where decorated with `@handle_klio`
* user-facing docs on where to view these metrics (both dataflow UI and stackdriver) (including changelog entry)

I'm having some issues with pickling with the counters & timer that I want to add to `@handle_klio` (I think it's just mostly unit tests, no pickling issues w/ dataflow/direct runner jobs), but want to at least get this reviewed for folks' feedback.

I've included unit tests, and tested with direct runner & dataflow both streaming and batch. 

To try out yourself, pull in these changes, build the `klio` wheel, and add it to your job's `Dockerfile` to be installed. Then run your job on dataflow, and you should see any relevant counters mentioned above. 🤞🏻 

## Checklist for PR author(s)

### DIFF
--- a/lib/src/klio/transforms/_helpers.py
+++ b/lib/src/klio/transforms/_helpers.py
@@ -171,6 +171,15 @@ class _KlioInputDataMixin(object):
 
     DIRECTION_PFX = KlioIODirection.INPUT
 
+    def setup(self, *args, **kwargs):
+        super(_KlioInputDataMixin, self).setup(*args, **kwargs)
+        self.found_ctr = self._klio.metrics.counter(
+            "kmsg-data-found-input", transform=self._transform_name
+        )
+        self.not_found_ctr = self._klio.metrics.counter(
+            "kmsg-data-not-found-input", transform=self._transform_name
+        )
+
     @property
     def _data_config(self):
         # TODO: figure out how to support multiple inputs
@@ -198,6 +207,11 @@ class _KlioInputDataMixin(object):
             )
         return self._klio.config.job_config.data.inputs[0]
 
+    @property
+    def _transform_name(self):
+        # grab the child class name that inherits this class, if any
+        return self.__class__.__name__
+
 
 class _KlioOutputDataMixin(object):
     """Mixin to add output-specific logic for a data existence check.
@@ -206,6 +220,15 @@ class _KlioOutputDataMixin(object):
     """
 
     DIRECTION_PFX = KlioIODirection.OUTPUT
+
+    def setup(self, *args, **kwargs):
+        super(_KlioOutputDataMixin, self).setup(*args, **kwargs)
+        self.found_ctr = self._klio.metrics.counter(
+            "kmsg-data-found-output", transform=self._transform_name
+        )
+        self.not_found_ctr = self._klio.metrics.counter(
+            "kmsg-data-not-found-output", transform=self._transform_name
+        )
 
     @property
     def _data_config(self):
@@ -235,6 +258,11 @@ class _KlioOutputDataMixin(object):
             )
         return self._klio.config.job_config.data.outputs[0]
 
+    @property
+    def _transform_name(self):
+        # grab the child class name that inherits this class, if any
+        return self.__class__.__name__
+
 
 class _KlioGcsDataExistsMixin(object):
     """Mixin for GCS-specific data existence check logic.
@@ -243,7 +271,8 @@ class _KlioGcsDataExistsMixin(object):
     _KlioInputDataMixin or _KlioOutputDataMixin
     """
 
-    def setup(self):
+    def setup(self, *args, **kwargs):
+        super(_KlioGcsDataExistsMixin, self).setup(*args, **kwargs)
         self.client = gcsio.GcsIO()
 
     def exists(self, path):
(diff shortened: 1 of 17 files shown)


lib/src/klio/transforms/io.py
```diff
@@ -109,14 +166,27 @@ def read_records(self, file_name, range_tracker):
...
     """Read from a local or GCS file with each new line as a
     ``KlioMessage.data.element``.
     """
 
     _source_class = _KlioReadFromTextSource
 
 
+class KlioReadFromText(beam.PTransform):
+    def __init__(self, *args, **kwargs):
+        self.reader = _KlioReadFromText(*args, **kwargs)
```
econchick, line 180: I made these readers/writers public/semiprivate so they can be tested easier. I just realized that I'm not consistent re: making public or semiprivate. I can change that if y'all want. 

I didn't need to mock out the counter objects so I kept those private.

lib/src/klio/transforms/_helpers.py
```diff
@@ -207,6 +218,17 @@ class _KlioOutputDataMixin(object):
 
     DIRECTION_PFX = KlioIODirection.OUTPUT
 
+    def setup(self, *args, **kwargs):
+        super(_KlioOutputDataMixin, self).setup(*args, **kwargs)
+        # grab the child class name that inherits this class
+        transform_name = self.__class__.__name__
```
DanSimon, line 224: wonder if we should just make `transform_name` a property on `_Klio(Input/output)DataMixin` since it's used in nearly every subclass now.

lib/src/klio/transforms/_helpers.py
```diff
@@ -260,9 +283,12 @@ def process(self, kmsg):
         item_path = self._get_absolute_path(item)
         item_exists = self.exists(item_path)
 
-        state = DataExistState.FOUND
         if not item_exists:
+            self.not_found_ctr.inc()
```
DanSimon, line 287: Is there any way that `_KlioOutputDataMixin` can handle the metrics increments?  Usually I prefer the pattern that whatever creates the counter also is responsible for incrementing it (though it's not too out-of-place here either).

DanSimon (approved):
couple small thoughts, otherwise LGTM 👍

lib/src/klio/transforms/_helpers.py
econchick: The `_KlioOutputDataMixin` doesn't have a `process` method; if we were to add a `process` method to increment counters in the Mixins (and therefore calling `super(...).process(kmsg)` here), it wouldn't know which counter to increment (found or not found), unless we were to move the entire logic of this `process` method up into both `_KlioOutputDataMixin` and `_KlioInputDataMixin`. I get your argument, just curious if you have a preference to follow the creator -> incrementor rule, or keep to DRY. I'm a bit ambivalent so opinions welcome 🙂

DanSimon: yeah I'm not sure there's a way to do it without making a bunch of other changes, so I'm fine with it as is.  It would be a bigger deal if we expected a lot of other subclasses or subclasses written outside of Klio, but I don't think either will be the case for a while.

econchick: fair enough 👍🏻 thanks!

PR #192, opened 2021-05-26
[cli] klio message publish blocks on pubsub client future

DanSimon:
When using `google-cloud-pubsub` `2.5.0`, `klio message publish` will say that the message is successfully published, but the message is not actually published to the topic.  This appears to have been some underlying threading change causing a race condition where `klio-cli` doesn't wait on the returned `Future` and exits before the message is actually published.

Simply blocking on the future fixes the issue.  I tested with pubsub `2.0.0`, `2.3.0` and `2.5.0`.

This is just a quick fix and there is room for further improvement.  If multiple messages are being published, this will end up blocking for each message, which is a bit inefficient since the pubsub client does have internal batching capabilities, but for small numbers of messages it shouldn't be big deal.

## Checklist for PR author(s)

### DIFF
--- a/cli/src/klio_cli/commands/message/publish.py
+++ b/cli/src/klio_cli/commands/message/publish.py
@@ -89,7 +89,9 @@ def _publish_messages(
             message = bytes(entity_id.encode("utf-8"))
 
         try:
-            publish(data=message)
+            future = publish(data=message)
+            # block until message is published or exception is raised
+            future.result()
             success_ids.append(entity_id)
 
         except Exception as e:
--- a/docs/src/faqs/performance.rst
+++ b/docs/src/faqs/performance.rst
@@ -1,4 +1,4 @@
 What's the performance of a Klio pipeline?
 ==========================================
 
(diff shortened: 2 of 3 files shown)


cli/src/klio_cli/commands/message/publish.py
```diff
@@ -89,7 +89,9 @@ def _publish_messages(
             message = bytes(entity_id.encode("utf-8"))
 
         try:
-            publish(data=message)
+            future = publish(data=message)
+            # block until message is published or exception is raised
+            future.result()
```
shireen-bean, line 94: Just double checking: I know this was related to a library update but is `future.result()` backwards compatible?

DanSimon: Yep, since we already require `google-cloud-pubsub>=2.0.0` and I tested successfully with `2.0.0`

shireen-bean (dismissed):
Mind adding that this was fixed in the changelog please?
So we can keep track of what's getting released in the fixes!
ty ty!!

DanSimon:
@shireen-bean wow I didn't even follow my own advice!  Added to the changelog, also added a commit to fix a docs link that apparently was just changed.

shireen-bean (approved):
Idk if this needed another ✅ before merging but thanks for adding the changelogs haha

PR #194, opened 2021-06-16
[lib] Add metrics interface for shumway

econchick:
Adds a user-facing metrics interface for [shumway](https://github.com/spotify/shumway) for when log-based metrics is not feasible. I skipped adding a note in `klio`'s changelog since I'm not sure what version this will be released in (alpha, beta, or release candidate).

* "I have included unit tests" 
* "I successfully ran my jobs with this code"

## Checklist for PR author(s)

### DIFF
--- a/docs/src/_templates/layout.html
+++ b/docs/src/_templates/layout.html
@@ -62,7 +62,7 @@
         {%- if meta is defined and meta is not none and 'nodocstoc' in meta %}
         {%- else %}
           {% block docs_toc %}
-              {% if not (meta is not none and 'notoc' in meta) %}
+              {% if meta is defined and not (meta is not none and 'notoc' in meta) %}
                 {%- include "docs-toc.html" %}
               {% endif %}
           {% endblock %}
--- a/docs/src/reference/lib/api/metrics/index.rst
+++ b/docs/src/reference/lib/api/metrics/index.rst
@@ -8,6 +8,7 @@
    Client <client>
    Native <native>
    Logger <logger>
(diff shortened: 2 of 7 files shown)


DanSimon (dismissed):
LGTM 👍 .  Only one question, what is the difference between attributes and tags, and do we want to let users specify custom attributes?

econchick:
@DanSimon good callout - Shumway has a concept of both `attributes` (a dict) and `tags` (a list of strings) ([ref](https://github.com/spotify/shumway/blob/master/shumway/__init__.py#L41-L54)). _(sidenote: it also has a concept of `resources` but that automatically gets populated by ffwd with hostname & podname, and maybe other resource-like things that may change on deployments.)_   But, I couldn't figure out how `tags` are handled with FFWD/Heroic (it _might_ have to do with the transition when we migrated away from riemann years ago). 

I didn't expose `attributes` to users since other metrics interfaces don't have that concept - they actually expose `tags` but as a dict instead of a list. And now, typing this out, I guess I should handle `tags` as a dict to be consistent with the metrics API, and just add it to `attributes`. And then not let users add the shumway-concept of `tags` as a list of strings since it's unclear how that gets handled or where it even appears. I'll fix that!

econchick:
@DanSimon kk - I updated `tags` to be a dict, so they have the same behavior as the rest of the metrics API, and effectively dropped Shumway's concept of `tags`. Mind a quick re-review?

DanSimon (approved):
👍

PR #197, opened 2021-06-17
[exec] Add  DirectGKERunner support

fallonchen:
We are implementing a new runner for Klio which facilitates running DIrectRunner on GKE, with assumption that this will only be used for streaming jobs. This PR includes changes to `klio-exec` needed to support `DirectGKERunner`.

This PR adds:

* `GkeDirectRunner` (and supporting classes). `GkeDirectRunner` inherits from `BundleBasedDirectRunner`, and  only differs from its implementation in that it overrides what we have been calling "autoacking" - `BundleBasedDirectRunner` pulls a message off the PubSub queue and immediately acknowledges it, while `GkeDirectRunner` extends the message acknowledgement deadline and waits until the Klio pipeline is done processing the message to acknowledge it. 
* `KlioAckInputMessage`, which is a DoFn that signals that processing in a Klio pipeline is done (so that `GkeDirectRunner` can acknowledge it)

This PR changes:

* `KlioAckInputMessage` is applied to the end of the KlioPipeline's transforms, but before event output messages are written. 

## Checklist for PR author(s)

- [] Document any relevant additions/changes in the appropriate spot in `docs/src`.
- [] For any change that affects users, update the package's changelog in ``docs/src/reference/<package>/changelog.rst``.


exec/src/klio_exec/runners/pubsub_message_manager.py
```diff
@@ -0,0 +1,284 @@
...
+        ] = KlioPubSubReadEvaluator
+
+
+class KlioAckInputMessage(beam.DoFn):
+    def process(self, element):
+        kmsg = klio_pb2.KlioMessage()
+        kmsg.ParseFromString(element)
+        entity_id = kmsg.data.element.decode("utf-8")
+
+        msg = ENTITY_ID_TO_ACK_ID.get(entity_id)
```
shireen-bean, line 383: I know you distinguish the Klio message as `kmsg` above but it did take a sec to realize `msg` is a `PubSubKlioMessage`. Totally non-blocking but it was helpful calling the variable `pmsg` like you did above since theres so many message references getting thrown around 😬

exec/src/klio_exec/runners/pubsub_message_manager.py
```diff
@@ -0,0 +1,284 @@
...
+class MessageManager:
+    DEFAULT_DEADLINE_EXTENSION = 30
+
+    def __init__(self, sub_name, heartbeat_sleep=10, manager_sleep=5):
+        self._client = g_pubsub.SubscriberClient()
+        self._sub_name = sub_name
+        self.heartbeat_sleep = heartbeat_sleep
+        self.manager_sleep = manager_sleep
+        self.messages = []
+        self.messages_lock = threading.Lock()
```
shireen-bean, line 86: Non-blocking just solidifying my understanding of whats going on:
So we have to make sure to have this messages lock because we have our main thread marking pmsgs in self.messages and pmsg.set() when its done processing, we have a heartbeat that accesses self.messages to give a heartbeat of what messages are still being processed as well as a MessageManager that actually does the `acking` of completed messages and thats the reason we have the lock?

econchick: not sure if I quite understand your understanding 😆 but maybe this will help: we use a lock in this case to prevent modifying the `self.messages` list (since other parts of the code could be adding to or removing from the list at any point) in the middle of us iterating over it in the `manage` method. If we didn't, we essentially get into a circumstance where the state of the list & iteration are unknown (was a message added to this list right after I finished the last loop? was a message deleted that I'm in the middle of handling?).  So, with this, we won't iterate over the messages if we are either adding to the list or removing from the list. And we won't add or remove from the list if we're currently iterating over it. Does that make sense?

shireen-bean: Perf, that makes sense. Thank you!

lib/src/klio/transforms/helpers.py
```diff
@@ -765,3 +766,20 @@ def expand(self, pcoll):
...
+
+
+# TODO: maybe expose a way for users to add messages to the pubsub message
+# manager as well? no idea if that'd be useful/needed
+class KlioAckInputMessage(beam.DoFn):
+    """Message acknowledgement DoFn.
+
+    Marks a KlioMessage as done after it is finished processing.
+
+    Used in conjunction with the ``klio_exec.runners.evaluators.
```
shireen-bean, line 778: Non-blocking. Might be worth adding that this only needs to be used with DirectGKERunner so someone doesnt use the Dataflow runner, wonder why the UI isnt acking messages as they expect thinking that they need this transform.

lib/src/klio/message/pubsub_message_manager.py
```diff
@@ -0,0 +1,312 @@
...
+            )
+        else:
+            self.mgr_logger.debug(
+                f"Extended Pub/Sub ack deadline for {message} by {duration}s"
+            )
+        message.extend(duration)
+
+    @staticmethod
+    def _convert_raw_pubsub_message(ack_id, pmessage):
+        # TODO: either use klio.mssage.serializer.to_klio_message, or
```
shireen-bean, line 213: Non-blocking. `mssage`. Theres a couple instances of it actually

lib/src/klio/transforms/decorators.py
```diff
@@ -160,6 +161,16 @@ def __get_thread_limiter(max_thread_count, thread_limiter, func_name=None):
     return thread_limiter
 
 
+def __ack_pubsub_if_direct_gke(kmsg_or_bytes, ctx):
+    # TODO: update me to `var.runners.DIRECT_GKE_RUNNER` once
```
shireen-bean, line 165: Non-blocking. This was changed to `var.KlioRunner.DIRECT_GKE_RUNNER`

exec/tests/unit/runners/test_evaluators.py
```diff
@@ -0,0 +1,381 @@
...
+
+
+@pytest.fixture
+def patch_msg_manager(mocker, monkeypatch):
+    p = mocker.Mock(name="patch_msg_manager")
+    monkeypatch.setattr(pmm, "MessageManager", p)
+    return p.return_value
+
+
+# @pytest.fixture
```
shireen-bean, line 54: Not sure if these are intentionally commented out!

shireen-bean (dismissed):
This LGTM.
The only thing i question is in the message timeout extension if the sleep manager 5 second sleep should be longer / less aggressive. In testing, most of my files were taking longer than 30 sec. Say around 45 so I would get like 7 or 8 "Skipping extending for message XYZ".
I know this is a personal use case where files took longer to process and this is a general media processing framework where we dont know how long processing will take. But just raising the thought because it came to mind but feel free to drop!

exec/src/klio_exec/commands/run.py
```diff
@@ -540,6 +557,21 @@ def _setup_pipeline(self, pipeline):
...
+            if to_pass_thru:
+                to_ack_input = (
+                    out_pcol,
+                    to_pass_thru,
+                ) | "Flatten to Ack Input Messages" >> beam.Flatten()
+            else:
+                to_ack_input = out_pcol
+
+            _ = to_ack_input | "Ack Input Messages" >> beam.ParDo(
+                helpers.KlioAckInputMessage()
```
DanSimon, line 572: Do we need to be acking messages ourselves if not using our custom direct runner?

It looks like when I run with regular direct runner, I get a `WARNING:klio.gke_direct_runner.message_manager:Unable to acknowledge test: Not found.` when it processes a message.  Haven't tried on dataflow yet.

DanSimon (approved):
Found one thing and I'm still doing some more testing, but overall looks good.

exec/src/klio_exec/commands/run.py
econchick: Hmm I'm confused, how does this line get hit with the `if` clause on 562?

DanSimon: ah nevermind I misread the line, and I think the warning I'm seeing was due to not having our other branch merged when I was testing.  I'll try again now that everything is merged together anyway

PR #198, opened 2021-06-17
[cli][core] Add DirectGKERunner as valid runner option and enable kubernetes interaction through the cli tools

shireen-bean:
## Changes

Add DirectGKERunner as valid config option for Klio
Add kubernetes interactions when using klio cli tools and runner is set to `DirectGKERunner`
Add respective unit tests for the above

## Testing
Unit tests added for the above functionality
Manually tested changes on this branch with changes in this PR https://github.com/spotify/klio/pull/197

## Checklist for PR author(s)

### DIFF
--- a/cli/src/klio_cli/cli.py
+++ b/cli/src/klio_cli/cli.py
@@ -135,9 +135,13 @@ def build_image(klio_config, config_meta, **kwargs):
 @options.runtime
 @core_utils.with_klio_config
 def run_job(klio_config, config_meta, **kwargs):
-    direct_runner = cli_utils.is_direct_runner(klio_config, kwargs.pop("direct_runner"))
+    direct_runner = cli_utils.is_direct_runner(
+        klio_config, kwargs.pop("direct_runner")
+    )
 
-    git_sha = cli_utils.get_git_sha(config_meta.job_dir, kwargs.get("image_tag"))
+    git_sha = cli_utils.get_git_sha(
+        config_meta.job_dir, kwargs.get("image_tag")
+    )
     image_tag = kwargs.get("image_tag") or git_sha
 
     runtime_config = DockerRuntimeConfig(
@@ -154,9 +158,12 @@ def run_job(klio_config, config_meta, **kwargs):
 
     if (
         not direct_runner
-        and klio_config.pipeline_options.runner == var.runners.DIRECT_GKE_RUNNER
+        and klio_config.pipeline_options.runner
+        == var.runners.DIRECT_GKE_RUNNER
     ):
-        run_gke = job_commands.gke.RunPipelineGKE(
+        from klio_cli.commands.job.gke import RunPipelineGKE
+
+        run_gke = RunPipelineGKE(
             config_meta.job_dir, klio_config, runtime_config, run_job_config
         )
         rc = run_gke.run()
@@ -170,7 +177,10 @@ def run_job(klio_config, config_meta, **kwargs):
 
 @job.command(
     "stop",
-    help=("Cancel a currently running job.\n\n**NOTE:** Draining is not " "supported."),
+    help=(
+        "Cancel a currently running job.\n\n**NOTE:** Draining is not "
+        "supported."
+    ),
 )
 @core_options.job_dir
 @core_options.config_file
@@ -191,7 +201,8 @@ def run_job(klio_config, config_meta, **kwargs):
 def stop_job(klio_config, config_meta, job_name, region, gcp_project):
     if job_name and any([config_meta.job_dir, config_meta.config_file]):
         logging.error(
-            "'--job-name' can not be used with '--config-file' and/or " "'--job-dir'."
+            "'--job-name' can not be used with '--config-file' and/or "
+            "'--job-dir'."
         )
         raise SystemExit(1)
 
@@ -210,9 +221,13 @@ def stop_job(klio_config, config_meta, job_name, region, gcp_project):
     # TODO: make this a click option once draining is supported @lynn
     strategy = "cancel"
     if klio_config.pipeline_options.runner == "DirectGKERunner":
-        job_commands.gke.StopPipelineGKE(config_meta.job_dir).stop()
+        from klio_cli.commands.job.gke import StopPipelineGKE
+
+        StopPipelineGKE(config_meta.job_dir).stop()
     else:
-        job_commands.stop.StopJob().stop(job_name, gcp_project, region, strategy)
+        job_commands.stop.StopJob().stop(
+            job_name, gcp_project, region, strategy
+        )
 
 
 @job.command(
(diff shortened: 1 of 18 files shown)


fallonchen (commented):
Just some comments on docstring formatting and log messages. The overall approach looks good. I will leave more feedback if anything turns up in manual testing!

cli/src/klio_cli/commands/job/gke.py
```diff
@@ -0,0 +1,218 @@
...
+
+
+class DeletePipelineGKE(GKECommandMixin):
+    def __init__(self, job_dir):
+        super().__init__()
+        self.job_dir = job_dir
+
+    def _delete_deployment(self):
+        dep = self.deployment_config
+        deployment_name = glom.glom(dep, "metadata.name")
```
econchick, line 197: Some places use `glom.glom` and some places use dictionary lookups for getting info from the deployment config. Could we be consistent?

cli/src/klio_cli/commands/job/gke.py
```diff
@@ -0,0 +1,218 @@
...
+    def __init__(self, job_dir):
+        super().__init__()
+        self.job_dir = job_dir
+
+    def stop(self):
+        """
+        Delete a namespaced deployment
+        Expects existence of a kubernetes/deployment.yaml
+        """
+        self._update_deployment(replica_count=0)
```
econchick, line 187: There seems to be an actual bug in this when the `deployment.yaml` has a different image tag than what was used when running `klio job run --image-tag $TAG`.

The steps/behavior I see:

1. Have two docker image versions, `my-image:1` and `my-image:2`.
2. In `deployment.yaml`, use the image `my-image:1`.
3. Run `klio job run --image-tag 2`
4. The deployment correctly runs with the `my-image:2` docker image.
5. Run `klio job stop`
6. The deployment first scales down to 0.
7. Unexpected: immediately after, the deployment then **scales up** to the original `Replicas` number with `my-image:1` as set in `deployment.yaml` file.

cli/src/klio_cli/commands/job/gke.py
```diff
@@ -0,0 +1,218 @@
...
+
+import glom
+import yaml
+from kubernetes import client, config
+
+from klio_cli.commands import base
+from klio_cli.utils import docker_utils
+
+
+class GKECommandMixin(object):
```
econchick, line 28: This seems to not use anything from the pipeline classes that it's mixed in with with the exception of `job_dir`. Does this need to be a mixin that's inherited? Could it just be a class representing/wrapping the k8s client? That gets instantiated with `job_dir` and provided to when one of the other pipeline classes are instantiated?

cli/src/klio_cli/commands/job/gke.py
```diff
@@ -0,0 +1,218 @@
...
+# limitations under the License.
+#
+
+import logging
+import os
+import re
+
+import glom
+import yaml
+from kubernetes import client, config
```
econchick, line 22: This goes against our style/the google style guide. It should be:

```py
from kubernetes import client
from kubernetes import config
```

Relatedly, could you name the imports something that makes it clear that it's a k8s client/config module? I got confused looking at `config` below thinking it was a klio-related config thing.

```py
from kubernetes import client as k8s_client
from kubernetes import config as k8s_config
```

(or something like that)

cli/src/klio_cli/commands/job/gke.py
```diff
@@ -0,0 +1,218 @@
+# Copyright 2019-2020 Spotify AB
```
econchick, line 1: It's 2021 :)

Also, all the `GKE` naming (even `runner: DirectGKERuner` in the job config) should probably be renamed to not make it sound like GKE is needed, as I would assume running a pipeline could work on AWS's EKS and elsewhere. What do y'all think?

cli/tests/commands/job/test_gke.py
```diff
@@ -0,0 +1,397 @@
...
+
+from klio_core import config
+
+from klio_cli import cli
+from klio_cli.commands.job import gke as job_gke
+
+
+@pytest.fixture
+def mock_os_environ(mocker):
+<<<<<<< HEAD
```
econchick, line 29: Looks like there are some git conflicts 😬

cli/tests/commands/job/test_gke.py
```diff
@@ -0,0 +1,397 @@
...
+    deployment_config, run_pipeline_gke, deployment_resp, monkeypatch, mocker
+):
+    deployment_name = deployment_config["metadata"]["name"]
+    namespace = deployment_config["metadata"]["namespace"]
+    mock_k8s_client = mocker.Mock()
+    mock_k8s_client.patch_namespaced_deployment.return_value = deployment_resp
+
+    # run_pipeline_gke = job_gke.RunPipelineGKE("/some/job/dir")
+
+<<<<<<< HEAD
```
econchick, line 253: More git conflicts 😬

cli/tests/commands/job/test_gke.py
```diff
@@ -0,0 +1,397 @@
...
+                    ],
+                },
+            },
+        },
+    }
+
+
+@pytest.fixture
+def deployment_resp():
+    # TODO: We can grab these from deployment_config
```
econchick, line 180: Should this be addressed now or is it too much work to bother with it now?

cli/src/klio_cli/cli.py
```diff
@@ -54,6 +55,7 @@
 DockerRuntimeConfig = collections.namedtuple(
     "DockerRuntimeConfig", ["image_tag", "force_build", "config_file_override"]
 )
+# TODO: We can get rid of direct_runner since we have runner
```
econchick, line 58: This seems like a relevant comment for `RuntimeConfig` in `exec/src/klio_exec/cli.py` too.

exec/src/klio_exec/cli.py
```diff
@@ -248,8 +246,8 @@ def profile_memory(
     "memory-per-line",
     short_help="Profile memory usage per line.",
     help=(
-        "Profile memory per line for every Klio-based transforms' process "
-        "method."
+        """Profile memory per line for every Klio-based
+         transforms' process method."""
```
econchick, line 250: Formatting looks weird here 😕 ❓

exec/src/klio_exec/cli.py
```diff
@@ -39,7 +39,8 @@
 
 
 RuntimeConfig = collections.namedtuple(
-    "RuntimeConfig", ["image_tag", "direct_runner", "update", "blocking"]
+    "RuntimeConfig",
```
econchick, line 42: I don't think any of the changes to the `klio-exec` code made in this PR are needed; it's taken care of in #197 . As well, we should not be telling the klio-exec to actually use the DirectGKERunner. By the time we are in this code path, we're already in the container on GKE. So we should be running using the plain direct runner.

econchick (changes requested):

cli/src/klio_cli/commands/job/gke.py
shireen-bean: I havent been able to recreate this issue. Just to confirm, the issue you're calling out here is the fact that it starts scaling up again despite having run `klio job stop` right? Not the image tag that we use?

The scaling up was definitely an issue in a previous iteration (longer ago than yesterday).

cli/src/klio_cli/commands/job/gke.py
```diff
@@ -0,0 +1,227 @@
...
+            if self.run_job_config.update:
+                self._update_deployment()
+            else:
+                logging.warning(
+                    f"Cannot apply deployment for {deployment_name}. "
+                    f"To update an existing deployment, run "
+                    f"`klio job run --update`, or set `pipeline_options.update`"
+                    f" to `True` in the job's`klio-job.yaml` file. "
+                    f"Run `klio job stop` to scale a deployment down to 0. "
+                    f"Run `klio job delete` to delete a deployment entirely."
```
econchick, line 162: FYI there's no need to add the `f` prefix for lines that don't have any variables (basically all but the first one). No need to change it here; but just wanted to surface it.

econchick (approved):
nice! ^5! :shipit:

econchick:
It looks like this branch hasn't been rebased on `develop` (in particular, it's missing [this commit](https://github.com/spotify/klio/pull/194/commits/9ac1ec5d4a48d21964eaec637cd2f3e564586b45) that passes the RTD build) - mind trying to rebase the branch again?

PR #203, opened 2021-07-20
[lib] Fix pubsub mgr with (1) being stuck in while loop and (2) missing msg extension deadline

econchick:
This PR contains two fixes/commits for two separate bugs seen while running direct on GKE. For both bugs, this was on python 3.8, apache beam 2.28, and klio* 21.7.0.dev2.

The first commit fixes a bug where the pubsub manager workers get stuck in the `while` loop and can't consume new messages. The behavior we've seen when testing this is that a message would be consumed, processed, and acked, but the heartbeat still continues (as seen with the log line `"Job is still processing <ID>"` every 10 seconds) and no new messages are consumed. How it gets into this state is unknown, but reproducible. This patch is more of a work-around where the heartbeat checks to see if the message is actually still in memory and if not, to ack (just in case) and break out of the loop.

The second commit fixes a bug related to extending the message deadlines. Currently, the pubsub manager consumes the message, and immediately extends the message deadline by 30 seconds (default). It then loops on the message, sleeps for 10 seconds, and checks to see if it should extend the message again (it extends if the last extension happened longer than 24 seconds (30 seconds * 80%)). So the first iteration is +10s since the message was last extended, the second iteration is +20s, and the third iteration is +30s, where it's too late. This caused a _lot_ of duplicate work, especially for the messages that take 30+ seconds to process. The change makes the sleep while looping shorter - to 3 seconds, so it was 1-2 opportunities to extend.

## Checklist for PR author(s)

### DIFF
--- a/lib/src/klio/message/pubsub_message_manager.py
+++ b/lib/src/klio/message/pubsub_message_manager.py
@@ -127,6 +127,21 @@ class MessageManager:
                 f"Job is still processing {message.kmsg_id}..."
             )
             time.sleep(self.heartbeat_sleep)
+            # Sometimes the heartbeat worker still continues in the `while`
+            # loop even though the message event flag has been set (not sure
+            # how this can happen...).
+            # So we're just checking to see if the message we're looping
+            # over still exists in ENTITY_ID_TO_ACK_ID; and if not, to
+            # go ahead and remove it. The `remove` method should be
+            # idempotent.
+            if message.kmsg_id not in ENTITY_ID_TO_ACK_ID:
+                self.hrt_logger.debug(
+                    f"No record of message {message.kmsg_id}. Assuming the "
+                    "pipeline is done processing it and will try to ack the "
+                    "message..."
+                )
+                self.remove(message)
+                break
 
     def _maybe_extend(self, message):
         """Check to see if message is done and extends deadline if not.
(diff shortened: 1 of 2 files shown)


shireen-bean (approved):
Cool this makes sense.
Also definitely not seeing the plateau of memory consumption after deploying with this patch.

lib/src/klio/message/pubsub_message_manager.py
```diff
@@ -127,6 +127,21 @@ def heartbeat(self, message):
...
+            # over still exists in ENTITY_ID_TO_ACK_ID; and if not, to
+            # go ahead and remove it. The `remove` method should be
+            # idempotent.
+            if message.kmsg_id not in ENTITY_ID_TO_ACK_ID:
+                self.hrt_logger.debug(
+                    f"No record of message {message.kmsg_id}. Assuming the "
+                    "pipeline is done processing it and will try to ack the "
+                    "message..."
+                )
+                self.remove(message)
```
DanSimon, line 143: If we're adding this in, is `message.event` even needed anymore?  Would it make more sense to just have `while message.kmsg_id in ENTITY_ID_TO_ACK_ID`?

DanSimon (approved):
I think the changes make sense as a workaround, although I've been looking into the details of using a `threading.Event`.  In most examples, it seems like this is used to wake up a sleeping thread and not necessarily to transmit information across two running threads (which will be the case here whenever `is_set()` is called).  Not sure if that could be leading to issues where the value of `is_set()` isn't correctly propagated between threads.

lib/src/klio/message/pubsub_message_manager.py
econchick: (responding to both this comment and your overall PR comment)

Good questions. So, I reached for thread event objects as a thread-safe means to control/signal (not `signal` signal) the two worker threads from the main thread (this [SO comment](https://stackoverflow.com/questions/18485098/python-threading-with-event-object#comment119832690_18485226) goes into detail). I also like using thread event objects because, at least to me, it signals (to the code reader) intent that, "this is a separate thread, and is dependent on another thread's actions." Or, at least that was _my_ intent. 

I'm happy to change it to be based on a dict lookup. I think I'd need to do a `while True`, and add a thread lock since the `ENTITY_ID_TO_ACK_ID` is meant to be modified (below there's `ENTITY_ID_TO_ACK_ID.pop(psk_msg.kmsg_id, None)`) and looping over a dict that may change in size can lead to unknown state.

lib/src/klio/message/pubsub_message_manager.py
```diff
@@ -109,38 +110,34 @@ def manage(self, message):
...
+                )
+                time.sleep(self.heartbeat_sleep)
+            else:
                 self.hrt_logger.debug(
-                    f"No record of message {message.kmsg_id}. Assuming the "
-                    "pipeline is done processing it and will try to ack the "
-                    "message..."
+                    f"Job is no longer processing {message.kmsg_id}."
                 )
-                self.remove(message)
```
econchick: I removed this call to `remove` here, hoping it's not needed. But I'm going to test this new commit before moving forward with merging anything to make sure.

lib/src/klio/message/pubsub_message_manager.py
DanSimon: Latest changes look good to me, using the lock makes sense too.  I think what I said with events was more of a guess and depends a lot on Python internals around multi-threading and caching.

On a similar note, how does Python handle accessing the same dict across threads, could there be a caching issue here too?  For reference, in Java you need to declare variables as `volatile` if you plan to write on one thread and read from another, since otherwise the write isn't always propagated across threads.  I have no idea if the same issues arise in Python.

DanSimon:
Going ahead with merging at @econchick 's request.  FYI, I amended a tiny change into the commit that changed formatting in a comment that was tripping up `black`

PR #206, opened 2021-08-06
[cli] Output GKE UI link whenever deployment updates

fallonchen:
Make it easier for users to access the GKE UI. 

## Checklist for PR author(s)

### DIFF
--- a/cli/src/klio_cli/commands/job/gke.py
+++ b/cli/src/klio_cli/commands/job/gke.py
@@ -184,7 +184,7 @@ class RunPipelineGKE(GKECommandMixin, base.BaseDockerizedPipeline):
                 body=dep, namespace=namespace
             )
             deployment_name = resp.metadata.name
-            current_cluster = self.kubernetes_active_context["name"]
+            current_cluster = self.kubernetes_active_context["name"][-1]
             ui_link = self._build_ui_link_from_current_context()
             logging.info(
                 f"Deployment created for {deployment_name} "
@@ -249,7 +249,7 @@ class DeletePipelineGKE(GKECommandMixin):
         deployment_name = glom.glom(dep, "metadata.name")
         namespace = glom.glom(dep, "metadata.namespace")
         # Some messaging might change if we multi-cluster deployments
-        current_cluster = self.kubernetes_active_context["name"]
+        current_cluster = self.kubernetes_active_context["name"][-1]
         if self._deployment_exists():
             resp = self.kubernetes_client.delete_namespaced_deployment(
                 name=deployment_name,
--- a/cli/src/klio_cli/commands/job/gke.py
+++ b/cli/src/klio_cli/commands/job/gke.py
@@ -184,11 +184,11 @@ class RunPipelineGKE(GKECommandMixin, base.BaseDockerizedPipeline):
                 body=dep, namespace=namespace
             )
             deployment_name = resp.metadata.name
-            current_cluster = self.kubernetes_active_context["name"][-1]
+            current_cluster = self.kubernetes_active_context["name"]
             ui_link = self._build_ui_link_from_current_context()
             logging.info(
                 f"Deployment created for {deployment_name} "
-                f"in cluster {current_cluster}"
+                f"in cluster {current_cluster}. "
                 f"Deployment details: {ui_link}"
             )
         else:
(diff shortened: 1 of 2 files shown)


cli/src/klio_cli/commands/job/gke.py
```diff
@@ -123,6 +129,26 @@ def _update_deployment(self, replica_count=None, image_tag=None):
...
             logging.info(message)
 
+        ui_link = self._build_ui_link_from_current_context()
+        logging.info(f"Deployment details: {ui_link}")
+
+    def _build_ui_link_from_current_context(self):
+        context_name = self.kubernetes_active_context["name"]
+        # context seems to follow the format
+        # gke_{project}_{region}_{clustername}
+        _, _, region, current_cluster = context_name.split("_")
```
fallonchen, line 139: Sticking to inferring the region for now rather than looking it up in the [GKE API](https://cloud.google.com/kubernetes-engine/docs/reference/rest/v1/projects.locations.clusters/list), since we have to parse the context anyway to get the name of the cluster.

cli/src/klio_cli/commands/job/gke.py
```diff
@@ -27,6 +27,12 @@
 
 
 class GKECommandMixin(object):
+    GKE_XPN_PROJECT = "gke-xpn-1"
+    GKE_UI_LINK_FORMAT = (
+        "https://console.cloud.google.com/kubernetes"
+        "/deployment/{region}/{cluster}/{namespace}/{app}"
+        "/overview?project=gke-xpn-1"
```
shireen-bean, line 34: Is this `project=gke-xpn-1` spotify specific or is that how all general users (including open source users) would access the k8s console?

fallonchen: Ah, good catch - I changed it to extract the project from the context rather than hardcoding it.

cli/src/klio_cli/commands/job/gke.py
```diff
@@ -184,11 +184,11 @@ def _apply_deployment(self):
                 body=dep, namespace=namespace
             )
             deployment_name = resp.metadata.name
-            current_cluster = self.kubernetes_active_context["name"][-1]
+            current_cluster = self.kubernetes_active_context["name"]
```
econchick, line 187: It looks like you undid the change in the previous commit - was this intentional? (similarly on line 252)

shireen-bean: yeah it was intentional. `self.kubernetes_active_context["name"][-1]` comes back empty for me vs `self.kubernetes_active_context["name"]` properly has the current cluster. Asking for some crowd sourcing to see if anyone can reproduce the need for the `[-1]` since users can only have one active context at a time.

econchick: ahhh gotcha. Here's a makeshift (non)repro:

```pycon
>>> from kubernetes import config as k8s_config
>>> _, active_context = k8s_config.list_kube_config_contexts()
>>> active_context["name"]
'gke_gke-xpn-1_europe-west1_testing-europe-west1'
```

Not sure where the indexing into the last element comes from either. I guess if you want to play it safe, you can do something like:

```py
current_cluster = self.kubernetes_active_context["name"]
if not isinstance(current_cluster, str):
  raise SomeException("can't detect current cluster name...blahblahblah")
```
or if we want to trust that, if it is a list, we should take the last element:

```py
current_cluster = self.kubernetes_active_context["name"]
if isinstance(current_cluster, list):
  current_cluster = current_cluster[-1]
```

Or we could not be defensive and just have it explode ¯\\_(ツ)_/¯ unless @fallonchen remembers how she got a list instead of a string. Without diving much into where [this gets set](https://github.com/kubernetes-client/python-base/blob/6b0104ffb9dd2f96d47d075ea3d30f69ea124ce4/config/kube_config.py#L598), it does look like it should be singular.

fallonchen: I'm not able to reproduce it either, which makes me think this was copy-pasta and that I was using an older version of my code when I tested it. I think it's safe to remove.  Sorry about the mess/confusion 😬 .

econchick (approved):
thank you @shireen-bean and @fallonchen ! 🚢

PR #210, opened 2021-08-13
[cli] Add k8s labels to deployments

econchick:
This ensures required and best-practices labels are set (which is particularly helpful for cost association), as well as any labels defined in the klio job config. This also adds a "deployed by" label and a "klio cli version" label, which can be particularly helpful during debugging (we currently have this for our Dataflow runner).

Since this code is not run inside the job's docker container, I wasn't able to get other klio package versions. While users can install mis-matching packages, hopefully with the recent change to calendar versioning + releasing-all-packages-per-release approach will minimize that. If worse comes to worst, we can add code to run something like `pip freeze | grep klio` in the job's container.

* I have included unit tests
* Instead of running a job, I dumped the resulting final `deployment.yaml` into a temp file with various initial `deployment.yaml` setups.
 
## Checklist for PR author(s)

### DIFF
--- a/cli/src/klio_cli/commands/job/gke.py
+++ b/cli/src/klio_cli/commands/job/gke.py
@@ -22,8 +22,24 @@ import yaml
 from kubernetes import client as k8s_client
 from kubernetes import config as k8s_config
 
+from klio_cli import __version__ as klio_cli_version
 from klio_cli.commands import base
 from klio_cli.utils import docker_utils
+
+
+# Regex according to https://kubernetes.io/docs/concepts/overview/
+# working-with-objects/labels/#syntax-and-character-set
+K8S_LABEL_KEY_PREFIX_REGEX = re.compile(
+    r"^([a-zA-Z0-9]|[a-zA-Z0-9][a-zA-Z0-9\-]{0,61}[a-zA-Z0-9])"
+    r"(\.([a-zA-Z0-9]|[a-zA-Z0-9][a-zA-Z0-9\-]{0,61}[a-zA-Z0-9]))*$"
+)
+K8S_LABEL_KEY_NAME_REGEX = re.compile(
+    r"^[a-zA-Z0-9]$|^[a-zA-Z0-9]([a-zA-Z0-9\._\-]){,61}[a-zA-Z0-9]$"
+)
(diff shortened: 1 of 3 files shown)


cli/src/klio_cli/commands/job/gke.py
```diff
@@ -160,6 +176,153 @@ def _apply_image_to_deployment_config(self):
...
+        # (these would be "deployment labels"). At least one label defined in
+        # `spec.template.metadata.labels` is required for k8s deployments
+        # ("pod labels").
+        # There also must be at least one "selector label"
+        # (`spec.selector.matchLabels`) which connects the deployment to pod.
+        # More info: https://stackoverflow.com/a/54854179
+        # TODO: add environment labels if/when we support dev/test/prod envs
+        metadata_labels, pod_labels, selector_labels = {}, {}, {}
+
+        # standard practice labels ("app" and "role")
```
econchick, line 247: "app" is standard practice for the k8s community in general (or a key similarly named). But "role" is a spotify standard practice/concept. I felt it was a bit of a hassle to separate it out into our internal spotify-specific plugins, and it shouldn't do any harm for non-spotify k8s-based jobs, so I just added it here. ¯\\_(ツ)_/¯

econchick:
FYI one of the integration tests - the audio-spectrograms - is failing because of (I assume temporary) PyPI caching issues. `matplotlib` released 3.4.3 yesterday, and for some reason, not all the built wheels are available on their CDNs ([matplotlib issue](https://github.com/matplotlib/matplotlib/issues/20834) links to [this PyPI issue](https://github.com/pypa/pypi-support/issues/1261)). Basically, the github action is downloading the `tar` file when it should download a particular prebuilt `whl`. The `tar` fails to install properly, causing the build issue that fails the test. Since it downloads the correct file when running this locally, I suspect the caching issues have not yet resolved.

I'll give it some time (i.e. the weekend) and rerun the integration test before this is merged in.

tacaswell:
This should be resolved now.

I also suggest setting the line in your requirements as

```
matplotlib --only-binary matplotlib
```

so if this happen again you will fall back to the newest existing wheel.

econchick:
Thanks @tacaswell ! much appreciated

fallonchen (approved):

PR #214, opened 2021-08-31
[cli] Respect all k8s yaml when deploying to DirectOnGKE

shireen-bean:
## Changes

klio-cli will now look in the kubernetes directory and respect all yaml files.

Workflow:

`klio job run`: Iterate through each *.yaml, if the resource exists it will warn a user that it already exists and they should use the `--update` flag; If the resource does not exist it will create the resource.

`klio job delete`: Iterate through each *.yaml, if the resource exists, it will be deleted

`klio job stop`: Find the first  *.yaml that contains kind=Deployment. This config will be used to scale the deployment to 0 replicas.

## Testing
Adjusted unit tests as well as tested manually

## Checklist for PR author(s)

### DIFF
--- a/cli/src/klio_cli/commands/job/gke.py
+++ b/cli/src/klio_cli/commands/job/gke.py
@@ -13,6 +13,7 @@
 # limitations under the License.
 #
 
+import glob
 import logging
 import os
 import re
@@ -21,6 +22,7 @@ import glom
 import yaml
 from kubernetes import client as k8s_client
 from kubernetes import config as k8s_config
+from kubernetes.dynamic.client import DynamicClient
 
 from klio_cli import __version__ as klio_cli_version
 from klio_cli.commands import base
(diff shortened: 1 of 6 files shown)


cli/src/klio_cli/commands/job/gke.py
```diff
@@ -52,110 +54,114 @@ class GKECommandMixin(object):
...
+            k8s_config.load_kube_config()
+            with k8s_client.ApiClient() as api_client:
+                dyn_client = DynamicClient(api_client)
+                self._dynamic_client = dyn_client
+        return self._dynamic_client
 
-    def _deployment_exists(self):
-        """Check to see if a deployment already exists
+    def get_deployment_config(self, config_dir="kubernetes"):
+        """Searches kubernetes directory for a Deployment
```
DanSimon, line 84: I wonder if (maybe in a subsequent PR) we should do something more comprehensive if there are actually multiple deployment files.  

I guess, does it make sense/is it not uncommon to have multiple such files?  If so, maybe we should require a specifier or be more explicit about which one is used by default.  If not, maybe we should error if multiple are detected.  Also maybe this is overkill so I'm also fine with not doing anything if others agree.

DanSimon (commented):
Looks good overall!  Just one minor comment.  Also, can you add a line to the changelog (trying to stay on top of that more!)

cli/src/klio_cli/commands/job/gke.py
shireen-bean: Yeah, for this PR i assumed that we deploy and delete the other resources like hpa and vpa but we interact with Deployments much more. which is why i made that `get_deployment_config` but if we find ourselves interacting with the other resources we can do something like `get_resource_config(resource_type="Deployment")`.
But leaning towards crossing that bridge when we see how we/users will interact with the other resources

cli/src/klio_cli/commands/job/gke.py
```diff
@@ -346,42 +383,46 @@ def _apply_labels_to_deployment_config(self):
...
             RunPipelineGKE._validate_labels(label_path, label_dict)
             glom.assign(
-                self.deployment_config, label_path, label_dict, missing=dict
+                deployment_config, label_path, label_dict, missing=dict
             )
 
-    def _apply_deployment(self):
-        """Create a namespaced deploy if the deployment does not already exist.
-        If the namespaced deployment already exists then
+    def _create_resource(self, resource_config):
```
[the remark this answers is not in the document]
shireen-bean: ooooh I'm thinking `apply_resource` to reflect some kubernetes language which does a create or an update!

DanSimon (approved):
💯

PR #218, opened 2021-09-07
[lib] replace config thread-local with global and thread lock

DanSimon:
When doing some benchmarking on streaming Dataflow, it was observed that if a job was started with large backlog of messages in Pubsub, the job would take a while to start processing messages and seem to have some throughput problems.

It was determined that the underlying cause was twofold.
1.  Dataflow will pull as many messages from pubsub at once, and Beam will attempt to create one thread per message with no upper bound.  Thus a large backlog can result in thousands of threads created simultaneously per worker machine.
2. Klio stores its cached config in a thread-local variable, which means each spawned thread must read the file from disk at least once.  Apparently when many threads are started at once, this causes some major I/O contention, which was the source of the lockup and low throughput.

Therefore, this PR replaces the thread-local storage with simply a single global variable guarded by a thread lock.  This means that now config is only read from disk once per worker and in a thread-safe manner.

This was tested extensively on Dataflow and with backlogs of over 100k messages, all startup problems were largely eliminated.  Beam will still start many threads but they no longer pile up loading config and processing begins shortly after the first thread is spawned.

Only one small issue: I added a simple unit test but due to how we mock out `RunConfig` in other tests, the test is affected and doesn't run properly.  I unsuccessfully made a couple quick attempts to fix, so we'll have to go back and figure out a better way to handle config mocking (or eliminate the need to begin with) if we want to not skip the test.

## Checklist for PR author(s)

### DIFF
--- a/docs/src/reference/lib/changelog.rst
+++ b/docs/src/reference/lib/changelog.rst
@@ -1,6 +1,21 @@
 Changelog
 =========
 
+.. _lib-21.9.0:
+
+21.9.0 (UNRELEASED)
+-------------------
+
+.. start-21.9.0
+
+Changed
+*******
+
+* Klio's config file is now read from disk once per worker instead of once per thread (See `PR 218 <https://github.com/spotify/klio/pull/218>`_).
+
+.. end-21.9.0
+
+
 .. _lib-21.8.0:
 
 21.8.0 (2021-09-03)
--- a/lib/tests/unit/transforms/test_core.py
+++ b/lib/tests/unit/transforms/test_core.py
@@ -161,6 +161,18 @@ def test_load_config_from_file_raises(config_dict, mocker, monkeypatch):
         core_transforms.RunConfig._load_config_from_file()
 
 
+def test_runconfig_load_once(mocker, monkeypatch):
+    # ensures config is only loaded once
+    mock_load = mocker.Mock()
+    monkeypatch.setattr(
+        core_transforms.RunConfig, "_load_config_from_file", mock_load
+    )
+    core_transforms.RunConfig._config = None
+    core_transforms.RunConfig.get()
+    core_transforms.RunConfig.get()
+    assert 1 == mock_load.call_count
+
+
 @pytest.mark.parametrize("thread_local_ret", (True, False))
 def test_job_property(thread_local_ret, mocker, monkeypatch):
     mock_func = mocker.Mock()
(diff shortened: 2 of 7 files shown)


lib/tests/unit/transforms/test_core.py
```diff
@@ -161,6 +161,23 @@ def test_load_config_from_file_raises(config_dict, mocker, monkeypatch):
         core_transforms.RunConfig._load_config_from_file()
 
 
+# FIXME: other test files (test_pubsub_message_manager.py and others) patch
```
shireen-bean, line 164: Just confirming: this is because our tests run in parallel and even resetting the value to None won't guarantee it will be empty when we call `.get()` if another test loads the config first?

shireen-bean:
The changes and explanations make sense to me! Not going to approve tho because would like multiple sets of eyes before the ✅

lib/tests/unit/transforms/test_core.py
econchick: @shireen-bean I _suspect_ (@DanSimon can correct me if I'm wrong) that this is because patches are "leaking" from other tests. FWIW our tests are not run in parallel.

DanSimon: I think I've gotten around the issue by adding some `patcher.stop()` calls.  I've updated the PR with my changes, tests all pass so it definitely works, though I'm open if there's another approach we should take.

docs/src/reference/lib/changelog.rst
```diff
@@ -1,6 +1,21 @@
...
+
+21.9.0 (UNRELEASED)
+-------------------
+
+.. start-21.9.0
+
+Changed
+*******
+
+* Klio's config file is now read from disk once per worker instead of once per thread (See `PR 218 <https://github.com/spotify/klio/pull/218>` _).
```
econchick, line 14: it looks like there's a space between the 2nd back tick and the underscore for the link

econchick: Ah yeah the link is broken: https://klio--218.org.readthedocs.build/en/218/reference/lib/changelog.html#changed

econchick (approved):
LGTM - just need to fix that link in the changelog

docs/src/reference/lib/changelog.rst
DanSimon: Fixed!

PR #220, opened 2021-09-30
[cli,lib] new opt-in setting to load config from main session

DanSimon:
Previously, we had attempted to serialize `KlioConfig` into the main session, which would be loaded by workers upon startup.  This was to be able to support CLI overrides and template values which wouldn't be present in the config file bundled in the docker image.  See #63.

We had to roll that back due to a bug in Beam that would not properly handle the situation if the pickled session file failed to download on the worker.  Thus some workers would randomly crash due to not having a config object to use.  See #147. This has been fixed in Beam 2.31.

This PR brings back the ability to load config from the main session with an opt-in config flag.  So in order for it to work, the setting must be set to `true` in the config file bundled in the image.  But once that file is loaded, Klio will then attempt to overwrite with the main session config if the setting is enabled.  When we had reverted this before, we left in the code to set the main session var, so this PR only has to change how config is loaded.

This was tested in several long-running jobs on Dataflow and no issues were found.  We know from when we tried to enable this before that the issue only showed up sporadically, but if an issue does occur, the setting can just be turned off and config will only be loaded from a file.

Once 2.31 is the earliest version of Beam we support (and once we've had time for more extensive testing), we can default this to `true` or simply remove the setting and only use main session.

## Checklist for PR author(s)


lib/src/klio/transforms/core.py
```diff
@@ -80,6 +84,8 @@ def get(cls):
         with cls._lock:
             if cls._config is None:
                 cls._config = cls._load_config_from_file()
+                if cls._config.job_config.use_main_session_config is True:
```
econchick, line 87: Is there a way to not need user config at all, and just do something (either at the top of the module or within this class):

```py
from apache_beam import __version__ as beam_version

# snip

use_main_session = False  # maybe this is a class var? 
major, minor, patch = beam_version.split(".") 
if int(major) == 2 and int(minor) >= 31:
    use_main_session = True
```

lib/src/klio/transforms/core.py
```diff
@@ -37,6 +37,10 @@ class RunConfig(object):
     _lock = threading.Lock()
     _config = None
 
+    @classmethod
```
econchick, line 40: Curious, why is this a class method? And not an attribute of the class (like `_lock` or `_config`)? Every time `_logger()` is called, it's a function call (`getLogger(...)`) when it could just be called once.

lib/src/klio/transforms/core.py
DanSimon: I think we could definitely do that eventually, maybe in `21.10.0`, but for now at least I want to keep this strictly opt-in just in case.

lib/src/klio/transforms/core.py
DanSimon: yeah not sure why I did that.  Fixed it!

lib/src/klio/transforms/core.py
econchick: Could you tell me a bit more as to why you want to keep this as strictly opt-in? Or maybe another way to ask it: why wouldn't users want to use_main_session if they're using a version that has the fixed bug?

What I'm concerned about is adding a new configuration setting, particularly if the plans are to deprecated in the following release.

Is it the case that we're not confident it will work/worried the session may eventually be lost? and therefore we want to have an "out" to let users toggle it off?

DanSimon: Yes, my concern is that we never definitively proved that the issues we ran into with using main session last time were in fact caused by the bug in Beam.  It obviously seems very likely the cause, but being unable to reliably reproduce the issue in testing also means I was unable to verify that the issue no longer exists.

Giving it some more thought though, it might just be better to enable it without having a config setting, and if someone actually runs into this issue again, we'd just push out a hot-fix reverting main session again, and whoever ran into it would just rollback to the previous release.

I'm hopeful that we can get more extensive testing done in a release candidate, so if we do have to flip back it happens before we push out the final release.

Lastly, I just checked and Beam is currently in release candidate for 2.33, which means if we time things right we can push this out without having to check Beam version, since the earliest version we'd support would be 2.31

econchick: Ok cool. I know we decided long ago that we'd only support the latest 3 versions of beam, but do users update their beam version along-side their Klio version? If you don't want to do the version check (which is fine), could we specify the minimum beam version instead?

DanSimon: > do users update their beam version along-side their Klio version?

I think there is a mix of some users not having a beam version specified in their `requirements.txt` and simply default to whatever Klio specifies (or what may already be included in their docker image), and other users having beam pinned to a specific version.

Either way, I think it's reasonable to recommend to users that they are always using a Beam version supported by the Klio version they're using.

> could we specify the minimum beam version instead?

We already have a min version in `setup.py` for `>=2.21.0` (which is very out of date anyway), so I can update that and update the tested versions in `tox.ini`.

It probably also makes sense to leave this unmerged until Beam 2.33 is released, unless we're ok temporarily only supporting 2 released beam versions, or targeting `2.33rc1` as the latest release

econchick: ok sounds good to me! ty!

PR #221, opened 2021-10-05
release 21.9.0rc1

DanSimon:
version bumps for `21.9.0rc1`

## Checklist for PR author(s)

### DIFF
--- a/exec/setup.py
+++ b/exec/setup.py
@@ -153,8 +153,8 @@ META_FILE = read(META_PATH)
 INSTALL_REQUIRES = [
     "attrs",
     "click",
-    "klio-core>=21.9.0.dev1",
-    "klio>=21.9.0.dev1",
+    "klio-core>=21.9.0rc1",
+    "klio>=21.9.0rc1",
     "pyyaml",
     # 2.22 added DirectRunner support for `DoFn.setup`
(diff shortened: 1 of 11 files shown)


exec/setup.py
```diff
@@ -153,8 +153,8 @@ def get_long_description(package_dir):
 INSTALL_REQUIRES = [
     "attrs",
     "click",
-    "klio-core>=21.9.0.dev1",
-    "klio>=21.9.0.dev1",
+    "klio-core>=21.9.0rc1",
```
shireen-bean, line 156: Is klio-exec dependent on 21.9.0rc1 for the other libraries because we want to keep the versions in sync even tho its not breaking to have like 21.8.0 for those libraries?
Fine either way just asking for future reference

DanSimon: Yeah I decided ultimately to keep them in sync for just simplicity.  Maybe we're hypothetically being too strict, but I don't think we're in a situation where users would have a real reason to use different versions of core/lib/exec.  I think it's also easier for us from a support perspective if we can assume projects are using synced versions.

That being said, if folks think it's better to do otherwise then I'm ok with it.  We'll just have to be a bit more vigilant about making sure we actually do bump the dependencies when it's required.

shireen-bean (approved):

PR #223, opened 2021-10-07
[CI] Add reusable build GH workflow

econchick:
This is the first PR of at least one other.

Overview of my thinking on how to approach this: Someone will create a `release-*` branch, update the changelogs, run `bumpversion`, and push to make a PR against `master`. Once merged into master, then the `.github/workflows/releasing/*` workflows will be kicked off that will do the building, testing installations, and uploading that's defined in `RELEASING.md`. 

This first workflow, `build.yml`, will be a "template workflow" (making use of GH's ["reusable workflow"](https://docs.github.com/en/actions/learn-github-actions/reusing-workflows) feature) that other workflows will call.

Once this commit is merged, the plan is to first test this build workflow with a manual trigger (like we can with our unit tests [here](https://github.com/spotify/klio/actions/workflows/manual.yml) via "Run workflow"; workflow defined [here](https://github.com/spotify/klio/blob/develop/.github/workflows/manual.yml)). 

Then once it works, I'll add another workflow that calls this workflow (which has to be committed in the repo for another workflow to call it) to do the releasing.

The second, forthcoming workflow will have a few jobs that depend on this `build.yaml` workflow. All jobs will build & locally test install the artifacts (this workflow), then upload to the test PyPI and test install, then finally upload to prod PyPI and test install. The first job will be for `klio-core`. The second job will be for `klio`, `klio-cli`, `klio-devtools`, and `klio-audio` and will be dependent on the first job. The third job will be for `klio-exec` since it's dependent on `klio-core` and `klio` being built and released.

I'll be sure to also update our `RELEASING.md` docs.

Definitely open to suggestions for different approaches and such!

## Checklist for PR author(s)

### DIFF
--- a/.github/workflows/releasing/build.yaml
+++ b/.github/workflows/releasing/build.yaml
@@ -1,0 +1,78 @@
+# "Template" workflow that can be called by other workflows as well as manually triggered
+# https://docs.github.com/en/actions/learn-github-actions/reusing-workflows
+
+name: Build Artifact & Test Local Installation
+
+on:
+  workflow_call:  # allows calls from other workflows
+    inputs:
+      package-dir:
+        description: "Directory of package to build"
+        required: true
+        type: string
+      package-name:
+        description: "Name of package to build"
+        required: true
+        type: string
+  workflow_dispatch:  # manual trigger
+    inputs:
+      package-dir:
+        description: "Directory of package to build"
+        required: true
+        type: string
+      package-name:
+        description: "Name of package to build"
+        required: true
+        type: string
+
+jobs:
+  build_install:
+    name: "Build & Install ${{ github.event.inputs.package-name }}"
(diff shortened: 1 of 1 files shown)


econchick:
@DanSimon ahh yeah that's a good point.

Re: tags - I'll have to dig into the docs a bit more, but it seems like `git push --tags ...` would trigger a workflow that's set to [`on: push` for tags](https://docs.github.com/en/actions/learn-github-actions/workflow-syntax-for-github-actions#onpushpull_requestbranchestags) (where if it was just `on: push` to `master`, it would include merging a PR into master), which is different than our workflow in that we'd have to merge the release branch to `develop` / `master` / `release-*`,  then push the tags.  (not sure if that point made sense...but I'll dig deeper into docs)

I'll have a think at how best to trigger for every type of release 🤔

PR #255, opened 2022-01-27
Add support for waiting until job reaches running state in streaming job

AndersonReyes:
When using streaming mode, the deployment can be successful but the pipeline could actually fail shortly after. By adding a flag to wait for the job to at least get in running state we could make the CI/CD deployment wait and error if the deployment was not successful. To avoid waiting infinitely, set some defaults 
```python
        timeout_sec=10 * 60,  # 10 minutes in seconds
        poll_interval_sec=60,
```

This way we only wait a max of 10 minutes, and only pool every minutes to check for pipeline state.

I havent made the timeout seconds and poll interval configurable yet until I get feedback from devs if this a good thing to do here at all or not.

I included unittest tests but also did a local test and pipeline eventually reaches done state
Screen Shot 2022-01-26 at 9 13 39 PM

-->

## Checklist for PR author(s)

### DIFF
--- a/core/src/klio_core/config/core.py
+++ b/core/src/klio_core/config/core.py
@@ -177,9 +177,11 @@ class KlioJobConfig(object):
     version = utils.field(type=int)
 
     # optional attributes
+
     allow_non_klio_messages = utils.field(type=bool, default=False)
     metrics = utils.field(default={})
     blocking = utils.field(type=bool, default=False)
+    wait_for_pipeline_running = utils.field(type=bool, default=False)
 
     def __config_post_init__(self, config_dict):
         self._raw = config_dict
--- a/exec/src/klio_exec/commands/run.py
+++ b/exec/src/klio_exec/commands/run.py
@@ -17,12 +17,12 @@ import collections
 import imp
 import logging
 import os
-import time
 import re
+import time
 
 import apache_beam as beam
-from apache_beam.runners.runner import PipelineState, PipelineResult
 from apache_beam.options import pipeline_options
+from apache_beam.runners.runner import PipelineResult, PipelineState
 
 from klio import __version__ as klio_lib_version
 from klio import transforms
@@ -565,9 +565,9 @@ class KlioPipeline(object):
         for _ in range(0, timeout_sec, poll_interval_sec):
             try:
                 status = result.state
-                if status in PipelineState.RUNNING:
+                if status == PipelineState.RUNNING:
                     logging.info(
-                        f"Pipeline status is %s, done waiting", status,
+                        "Pipeline status is %s, done waiting", status,
                     )
                     return status
                 elif PipelineState.is_terminal(status):
@@ -575,7 +575,7 @@ class KlioPipeline(object):
                     return status
 
                 logging.info(
-                    f"Pipeline status %s not in %s, retrying in %s seconds",
+                    "Pipeline status %s not in %s, retrying in %s seconds",
                     status,
                     PipelineState.RUNNING,
                     poll_interval_sec,
@@ -586,7 +586,8 @@ class KlioPipeline(object):
             time.sleep(poll_interval_sec)
 
         raise TimeoutError(
-            f"Pipeline finished in status {status} but expected {PipelineState.RUNNING}"
+            f"Pipeline finished in status {status}"
+            f"but expected {PipelineState.RUNNING}"
         )
 
     def run(self):
@@ -636,7 +637,6 @@ class KlioPipeline(object):
             result.wait_until_finish()
 
         # If the blocking flag was already passed don't wait again
-        # import pdb; pdb.set_trace()
         if (
             self.config.job_config.wait_for_pipeline_running
             and self.config.pipeline_options.streaming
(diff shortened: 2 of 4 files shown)


exec/src/klio_exec/commands/run.py
```diff
@@ -552,6 +554,41 @@ def _setup_pipeline(self, pipeline):
...
+        result: PipelineResult,
+        timeout_sec=10 * 60,  # 10 minutes in seconds
+        poll_interval_sec=60,
+    ):
+        status = None
+
+        for _ in range(0, timeout_sec, poll_interval_sec):
+            try:
+                status = result.state
+                if status in PipelineState.RUNNING:
```
AndersonReyes, line 568: That’s supposed to be status = PipelineState.RUNNING but I guess it would still work but I’ll fix this.

exec/src/klio_exec/commands/run.py
```diff
@@ -586,11 +623,23 @@ def run(self):
...
+            )
         )
-        if any(should_block):
+
+        if should_block:
             # the pipeline on direct runner will otherwise get garbage collected
             result.wait_until_finish()
+
+        # If the blocking flag was already passed don't wait again
+        # import pdb; pdb.set_trace()
```
AndersonReyes, line 639: Remove

exec/src/klio_exec/commands/run.py
```diff
@@ -18,9 +18,11 @@
 import logging
 import os
 import re
+import time
 
 import apache_beam as beam
 from apache_beam.options import pipeline_options
+from apache_beam.runners.runner import PipelineResult, PipelineState
```
econchick, line 25: Nitpick: we follow the google style of code, which means we import modules, not specific classes/objects from modules. Mind changing this to `from apache_beam.runners import runner`?

exec/src/klio_exec/commands/run.py
```diff
@@ -552,6 +554,42 @@ def _setup_pipeline(self, pipeline):
                     **output_config.to_io_kwargs()
                 )
 
+    def wait_for_pipeline_running(
+        self,
+        result: PipelineResult,
```
econchick, line 559: I 100% agree with the use of typing. However, Klio does not use typing anywhere, so this would be an anomaly. I'm fine with just `result`, but if you want, maybe renaming this arg to `pipeline_result` would do to help with the lack of typing?

exec/src/klio_exec/commands/run.py
```diff
@@ -552,6 +554,42 @@ def _setup_pipeline(self, pipeline):
...
+        poll_interval_sec=60,
+    ):
+        status = None
+
+        for _ in range(0, timeout_sec, poll_interval_sec):
+            try:
+                status = result.state
+                if status == PipelineState.RUNNING:
+                    logging.info(
+                        "Pipeline status is %s, done waiting", status,
```
econchick, line 570: Nitpick - there's mixing between f-strings and string interpolation for these logs. Mind sticking to f-strings (or str.format)?

core/src/klio_core/config/core.py
```diff
@@ -177,9 +177,11 @@ class KlioJobConfig(object):
     version = utils.field(type=int)
 
     # optional attributes
+
     allow_non_klio_messages = utils.field(type=bool, default=False)
     metrics = utils.field(default={})
     blocking = utils.field(type=bool, default=False)
+    wait_for_pipeline_running = utils.field(type=bool, default=False)
```
econchick, line 184: Mind adding just a little blurb [here in our config docs](https://github.com/spotify/klio/blob/develop/docs/src/userguide/config/job_config.rst) about this value? Just so that users know it's available to configure.

econchick (changes requested):
Overall I like this idea - I think it's good to get it in. All the feedback is more nitpick-y 😅

exec/src/klio_exec/commands/run.py
AndersonReyes: 👍🏾

exec/src/klio_exec/commands/run.py
AndersonReyes: 👍🏾

core/src/klio_core/config/core.py
AndersonReyes: 👍🏾

exec/tests/unit/commands/test_run.py
```diff
@@ -644,7 +644,9 @@ def test_get_pipeline_options(
 )
 @pytest.mark.parametrize("blocking", (True, False))
 @pytest.mark.parametrize("streaming", (True, False))
+@pytest.mark.parametrize("wait_until_pipeline_complete", (True, False))
```
AndersonReyes, line 647: This should be renamed to pipeline running. Causes confusion

shireen-bean:
Closing this PR due to these changes being merged in via this PR https://github.com/spotify/klio/pull/263

AndersonReyes:
@econchick @shireen-bean yeah thank you fixing! Forgot to come back to this.

